PR #23, opened 2015-01-08
Introduce DecodedType typealias

gfontenot:
Introducing this typealias helps when trying to conform a non-final class to
`JSONDecodable`.

If we just use `Self`, we run into an issue where the compiler can't ensure
that return type is correct, because the class might be subclassed. By letting
classes explicitly define their decoded type, we can make the compiler happy
while maintaining the ability to subclass if that's what we want.

As an added benefit, we can default `DecodedType` to `Self`, which means that
`struct`s and `final` classes don't need to worry about changing anything from
the way they are doing it now.

Note that I'm adding a `JSONDecodable` version of `NSURL` for the tests.
This might be something that's common enough to warrant pulling into the
actual source in the future.

### DIFF
--- a/Argo/Operators/JSONOperators.swift
+++ b/Argo/Operators/JSONOperators.swift
@@ -1,7 +1,7 @@
 // MARK: Values
 
 // Pull embedded value from JSON
-public func <|<A: JSONDecodable>(json: JSONValue, keys: [String]) -> A? {
+public func <|<A where A: JSONDecodable, A == A.DecodedType>(json: JSONValue, keys: [String]) -> A? {
   if let o = json.find(keys) {
     return A.decode(o)
   }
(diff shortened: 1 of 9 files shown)


Argo/Operators/JSONOperators.swift
```diff
@@ -10,39 +10,39 @@ public func <|<A: JSONDecodable>(json: JSONValue, keys: [String]) -> A? {
...
 
 // Pull embedded array from JSON
-public func <||<A: JSONDecodable>(json: JSONValue, keys: [String]) -> [A]? {
+public func <||<A where A: JSONDecodable, A == A.DecodedType>(json: JSONValue, keys: [String]) -> [A]? {
   return json.find(keys) >>- JSONValue.map
 }
 
 // Pull array from JSON
-public func <||<A: JSONDecodable>(json: JSONValue, key: String) -> [A]? {
+public func <||<A where A: JSONDecodable, A.DecodedType == A>(json: JSONValue, key: String) -> [A]? {
```
swizzlr: Style: this is inconsistent with every other type constraint. Suggest you change to `A == A.DecodedType` for readability.

gfontenot: Good catch. Fixed.

swizzlr:
This looks good to me, I think it's worth adding a note for the record here that this is ultimately due to the problem described and answered here: http://stackoverflow.com/questions/25645090/protocol-func-returning-self.

tonyd256:
:+1: Make sure to bump the version to v0.2.2 and update the podspec file

PR #24, opened 2015-01-08
Extract functional operators to external framework

gfontenot:
Including these directly in Argo could lead to compilation issues if they were
also defined somewhere else. Instead, it probably makes more sense to extract
them to a common dependency, and then rely on them. This way, other users can
simply import the same framework and we don't have to worry about conflicts.

Note that since we're including this new dependency, we now have to check in
the `Carthage/Checkouts` folder in order to get everything to build properly
in client applications.

We will also need to add Runes to CocoaPods in order to be able to bump the
Podspec version.


tonyd256:
:+1:

tonyd256:
:+1:

PR #27, opened 2015-01-09
WIP: Improve README and documentation

gfontenot:
I'd like to try to improve the accessibility of this framework by
putting more effort into educating people about the concepts and
ideologies we're using. It'd like to take an initial stab at that by
improving the documentation and language used in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,54 +35,117 @@ pod 'Argo', :git => 'https://github.com/thoughtbot/Argo'
 
 Then run `pod install` with CocoaPods 0.36 or newer.
 
-### Git Submodules ###
+### Git Submodules
 
 I guess you could do it this way if that's your thing.
 
 Add this repo as a submodule, and add the project file to your workspace. You
 can then link against `Argo.framework` for your application target.
 
-## Usage
-
-First, create your model. We like to use structs but a class is OK too.
+## Usage tl;dr:
 
 ```swift
 struct User {
   let id: Int
   let name: String
   let email: String?
+  let role: Role
+  let companyName: String
+  let friends: [User]
 }
-```
 
-Then, extend the model to conform to `JSONDecodable`. You will also need to
-write a static constructor method that is curried. We like to use `create`.
-
-```swift
 extension User: JSONDecodable {
-  static func create(id: Int)(name: String)(email: String?) -> User {
-    return User(id: id, name: name, email: email)
+  static func create(id: Int)(name: String)(email: String?)(role: Role)(companyName: String)(friends: [User]) -> User {
+    return User(id: id, name: name, email: email, role: role, companyName: companyName, friends: friends)
+  }
+
+  static func decode(j: JSONValue) -> User? {
+    return User.create
+      <^> j <| "id"
+      <*> j <| "name"
+      <*> j <|? "email" // Use ? for parsing optional values
+      <*> j <| "role" // Custom types that also conform to JSONDecodable just work
+      <*> j <| ["company", "name"] // Parse nested objects
+      <*> j <|| "friends" // parse arrays of objects
+  }
+}
+
+// Wherever you receive JSON data:
+
+let json: AnyObject? = NSJSONSerialization.JSONObjectWithData(data, options: NSJSONReadingOptions(0), error: .None)
+
+if let j: AnyObject = json {
+  if let value = JSONValue.parse(j) {
+    let user = User.decode(value)
   }
 }
 ```
 
-It's important that the `create` function's parameters are all in separate
-parenthesis. This will make the function curried. Then inside the function we
-can call the model's constructor.
+## Ideology
 
-Finally, implement the `JSONDecodable` `decode` function to decode the incoming
-JSON. Here is an example of decoding a `User` from JSON received from a network
-request.
+Argo's core concept is that in order to maintain type safety, you should only
+be able to successfully decode an object if all parameters are satisfied
+properly. So if you have a model that looks like this:
 
-JSON received from network request:
-```
-{
-  "id": 1,
-  "name": "Cool User",
-  "email": "cool.user@example.com"
+```swift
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,54 +35,107 @@ pod 'Argo', :git => 'https://github.com/thoughtbot/Argo'
...
+
+If you aren't familiar with how these functions work (or just aren't
+comfortable with using operators), that's totally OK. It's possible to use the
+library without using them, although it might be a little more painful.
+
+If you're looking to learn more about these functions, we would recommend
+reading the following articles:
+
+- [Functional Swift for Dealing with Optional Values](http://robots.thoughtbot.com/functional-swift-for-dealing-with-optional-values)
+- [Railway Oriented Programming](http://fsharpforfunandprofit.com/posts/recipe-part2/)
```
gfontenot: Any other articles I should point out?

README.md
```diff
@@ -94,134 +147,145 @@ if let j: AnyObject = json {
...
-{
-  "id": 5,
-  "text": "A cool story.",
-  "author": {
-    "id": 1,
-    "name": "Cool User"
-  }
-}
-```
+- [Introduction to Function Currying in Swift](http://robots.thoughtbot.com/introduction-to-function-currying-in-swift)
```
gfontenot: Any other articles I should point out?

README.md
```diff
@@ -35,54 +35,107 @@ pod 'Argo', :git => 'https://github.com/thoughtbot/Argo'
...
+  static func create(id: Int)(name: String)(email: String?)(role: Role)(friends: [User]) -> User {
+    return User(id: id, name: name, email: email, role: role, friends: friends)
+  }
+
+  static func decode(j: JSONValue) -> User? {
+    return User.create
+      <^> j <| "id"
+      <*> j <| "name"
+      <*> j <|? "email" // Use ? for parsing optional values
+      <*> j <| ["role", "title"] // parse nested attributes
```
tonyd256: role looks like it should be the model `Role` but here you're pulling out the `title` which I'm assuming is a `String`. would it be better to have `roleTitle` as the property name?

README.md
```diff
@@ -35,54 +35,107 @@ pod 'Argo', :git => 'https://github.com/thoughtbot/Argo'
...
+      <*> j <|| "friends" // parse arrays of objects
+  }
+}
+
+// Wherever you receive JSON data:
+
+let json: AnyObject? = NSJSONSerialization.JSONObjectWithData(data, options: NSJSONReadingOptions(0), error: .None)
```
jferris: Have you thought about encapsulating this in a function like `class func parse(String) -> DecodedType?`?

README.md
```diff
@@ -35,54 +35,107 @@ pod 'Argo', :git => 'https://github.com/thoughtbot/Argo'
...
 
-Parse data to JSON then to a `JSONValue` and finally to our `User` object:
+you would want that JSON parsing to fail and return `.None` instead of a
+`User` object. That base concept means that when you're dealing with an actual
+`User` object, you can be sure that when you ask for `user.id`, you're going
+to get the type you expect.
+
+## Functional Concepts
+
+Argo really wants to be used with patterns stolen from Functional Programming
```
tonyd256: nit pick: stolen feels to negative .. how about "seen in" instead of "stolen from" or "used in"

README.md
gfontenot: We had discussed wrapping up the `NSData -> JSONValue` part into something. I think we'd still want to keep the `JSONValue -> DecodedType?` part separate though.

README.md
```diff
@@ -94,134 +147,145 @@ if let j: AnyObject = json {
...
 }
+```
 
-extension Post: JSONDecodable {
-  static func create(id: Int)(text: String)(author: User) -> Post {
-    return Post(id: id, text: text, author: author)
-  }
+You will also want to create a curried creation function. This is needed
+because of a deficiency in Swift that doesn't allow us to pass normal `init`
+functions around as normal functions.
```
tonyd256: normal is used twice ... this line and the one above ... I think getting rid of the one on the line above is better ... "to pass `init` functions..."

segiddins:
Inline [documentation comments](http://nshipster.com/swift-documentation/) would also be excellent!

README.md
```diff
@@ -94,134 +147,145 @@ if let j: AnyObject = json {
...
+```swift
+extension User {
+  static func create(id: Int)(name: String) -> User {
+    return User(id: id, name: name)
   }
 }
 ```
 
-From the JSON:
+Using this curried syntax will allow us to partially apply the function over
```
tonyd256: should partial application be explained more too?

README.md
tonyd256: we could reference the objc snippets if they have any on this stuff

README.md
gfontenot: I think that's explained in the linked article, but I can make it more explicit that it's explained there.

README.md
gfontenot: Yeah, this wasn't great. I wanted an example of a nested attribute, but also wanted an example of a custom type. I can break it into two different parameters.

README.md
```diff
@@ -35,54 +35,109 @@ pod 'Argo', :git => 'https://github.com/thoughtbot/Argo'
...
+you would want that JSON parsing to fail and return `.None` instead of a
+`User` object. That base concept means that when you're dealing with an actual
+`User` object, you can be sure that when you ask for `user.id`, you're going
+to get the type you expect.
+
+## Functional Concepts
+
+Argo really wants to be used with patterns borrowed from functional programming
+such as `map` and `apply`. We feel that these patterns greatly reduce the pain
+felt in trying to use JSON (an inherently loosely typed language) with Swift
```
hyperspacemark: I wouldn't call JSON a language. It's a format.

PR #21, opened 2015-01-08
Update podspec to support iOS 7 and latest version

tonyd256:

PR #29, opened 2015-01-12
Only support iOS 8+ with CocoaPods

gfontenot:
As noted in #21, frameworks are an iOS 8+ feature, so setting this to 7.0
doesn't really make sense. We should require 8.0 when installing through
CocoaPods.


tonyd256:
:+1:

PR #30, opened 2015-01-12
Misc cleanup tasks

gfontenot:
Just fixing up a few things stylistically that were bothering me.

### DIFF
--- a/Argo/Globals/JSONValue.swift
+++ b/Argo/Globals/JSONValue.swift
@@ -1,7 +1,9 @@
 import Foundation
 
+public typealias JSONDict = [String: JSONValue]
+
 public enum JSONValue {
-  case JSONObject([String:JSONValue])
+  case JSONObject(JSONDict)
   case JSONArray([JSONValue])
   case JSONString(String)
   case JSONNumber(NSNumber)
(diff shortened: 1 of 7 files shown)


Argo/Globals/JSONValue.swift
```diff
@@ -1,24 +1,24 @@
...
 
-    case let v as [String:AnyObject]:
-      var object: [String:JSONValue] = [:]
-      for key in v.keys {
-        if let value: AnyObject = v[key] {
-          object[key] = parse(value)
-        } else {
-          object[key] = .JSONNull
-        }
+    case let v as [String: AnyObject]:
```
tonyd256: `as JSONDict` maybe? but i can also see you liking the full type here ... either way I'm ok

gfontenot: `JSONDict` is `[String: JSONValue]`. This is `[String: AnyObject]`. I think it's actually just an `NSDictionary` that's being bridged to `Dictionary`.

tonyd256: yup right nvm

tonyd256:
:+1:

PR #32, opened 2015-01-14
Fix bad access in JSONValue.parse

aschuch:
Using the following JSON, `JSONValue.parse` would just throw a bad access.

```json
[
  {
    "title": "Foo",
    "age": 21
  },
  {
    "title": "Bar",
    "age": 32
  }
]
```

This makes the `[String: AnyObject]` branch of `parse` work again.

### DIFF
--- a/Argo/Globals/JSONValue.swift
+++ b/Argo/Globals/JSONValue.swift
@@ -15,14 +15,14 @@ public extension JSONValue {
   static func parse(json: AnyObject) -> JSONValue {
     switch json {
     case let v as [AnyObject]: return .JSONArray(v.map(parse))
-      
+
     case let v as [String: AnyObject]:
       let object = reduce(v.keys, JSONDict()) { accum, key in
         let value = self.parse(v[key]!)
         return Dictionary.appendKey(accum, key: key, value: value)
       }
       return .JSONObject(object)
-      
+
     case let v as String: return .JSONString(v)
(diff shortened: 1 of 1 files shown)


swizzlr:
What causes the bad access?

Argo/Globals/JSONValue.swift
```diff
@@ -18,8 +18,8 @@ public extension JSONValue {
 
     case let v as [String: AnyObject]:
       let object = reduce(v.keys, JSONDict()) { accum, key in
-        let append = curry(Dictionary.appendKey)(accum)(key)
-        return (append <^> v[key] >>- self.parse) ?? append(.JSONNull)
+        let value = self.parse(v[key]!)
```
tonyd256: We should never explicitly unwrap anything for safety reasons. If the bind (`>>-`) is causing bad access then we should try and replace it with an equivalent that does work.

gfontenot:
I'm like 99% sure that the bad access you're seeing is because of using `flatMap` inappropriately. I've opened #33 as a fix and included a test for your example JSON (although I couldn't replicate a failure there. Most of these crashes are really specific to the project).

aschuch:
Thanks. The pull request was just meant to give a hint to my problem rather than a definite fix. 
I have successfully tested with #33 and it is working again, so I'm closing this.

gfontenot:
@aschuch Awesome, thanks for reporting it. I've merged the fix and bumped the version.

PR #33, opened 2015-01-14
Use `map` instead of `flatMap` when parsing

gfontenot:
This was my bad, so I'm fixing it.

`JSONValue.parse` returns a non-optional `JSONValue`. Unfortunately, because
of Swift Magic, Xcode will happily compile when using `flatMap` with a
function that returns a non-optional. This can (but doesn't always) result in
a EXC_BAD_ACCESS error at runtime. Using `map` here is actually the correct
behavior, but should also fix this crash that people were seeing.

Fixes #32 (Or, it should).


aschuch:
:+1:

tonyd256:
:+1:

PR #35, opened 2015-01-15
Use Release scheme for tests and make them pass

gfontenot:
It's embarrassing how many times we've made this `bind`/`fmap` mistake.
I really wish Swift would prevent this from compiling in the first
place.


tonyd256:
:+1:

PR #36, opened 2015-01-15
Use `pure` from Runes

gfontenot:
- Update Runes to 1.1
- Remove our implementations of `pure` in favor of the implementation
  that comes with Runes


tonyd256:
:+1:

PR #43, opened 2015-02-10
Use static func instead of class func

klaaspieter:
The only change needed for Swift 1.2 support.


tonyd256:
Thanks for the PR. I already had a branch open with the changes. Please refer to #44

tonyd256:
@klaaspieter version 0.4.0 should take care of this now.

PR #44, opened 2015-02-10
Update for Swift 1.2

tonyd256:
This updates Argo to be compatible with Swift 1.2b. The only changed needed was `class` to `struct` in the `JSONDecode` protocol `decode` function.

Also, with the update came many stability fixes allowing us to partially apply functions from generic protocol static functions.


dazmuda:
Glad you get to use the runes operators now! Once the tests pass, this looks good.

tonyd256:
The tests won't pass because CircleCI is using Xcode 6.1 and won't switch to 6.3 until it's a stable release.

kaishin:
This looks good. I'd have split it into two, but oh well.

PR #46, opened 2015-02-11
Enable app extension API compatibility

kaishin:
Without this option, Xcode will throw a warning when the framework is used in
an app extension:

`linking against dylib not safe for use in application extensions`

As long as the framework is not using any APIs marked as
`NS_EXTENSION_UNAVAILABLE_IOS/MAC`, this should not affect anything.


gfontenot:
:+1:

Issue #42, opened 2015-02-03
Add JSONEncodable as well?


nvh:
I don't know if it's seen as in the scope of this project, but wouldn't it be nice to have a JSONEncodable protocol as well? Something like this:
```swift
public protocol JSONEncodable {
  func encode() -> JSONValue
}
```

With some smart use of the LiteralConvertible protocols the implementation could look something like this:
```swift
extension User: JSONEncodable {
  func encode() -> JSONValue {
    return ["id":id,"name":name,"email":email]
  }
}
```
I've made an example of this in my fork of Argo: https://github.com/nvh/Argo/blob/master/Argo/Globals/JSONValue.swift#L95

Will this be something worth making a pull request from?

tonyd256:
@nvh what have you been using this for? I have been thinking about adding this but not to a `JSONValue`. Are you finding benefit in having the JSONValue type? I was thinking about adding it as such: `func encode() -> AnyObject`. This way you can return the dictionary or whatever directly.

it would still look like this:
```
extension User: JSONEncodable {
  func encode() -> AnyObject {
    return ["id":id,"name":name,"email":email]
  }
}
```

nvh:
@tonyd256 Well, if you return AnyObject, you loose all type information. So when transforming to a json string, you have to figure out all the types again.

When returning JSONValue, you could do something like:

```swift
extension JSON {
  public func toString() -> String {
    switch(self) {
    case let .string(string):
      return "\"\(string)\""
    case let .number(number):
      return "\(number)"
    case let .object(object):
      let result = join(",",map(object,{ (key,value) in "\"\(key)\":\(value.description)"}))
      return "{\(result)}"
    case let .array(array):
      let result = join(",",map(array,{$0.description}))
      return "[\(result)]"
    case let .boolean(boolean):
      return boolean ? "true" : "false"
    case .null:
      return "null"
    }
  }
}
``` 

And have it return a valid json string. 
I know the above example doesn't map 100% to the current JSONValue type in Argo (this example is taken from one of my own projects). But it's a really nice way of recursively constructing JSON

tonyd256:
I don't understand why you need to convert it to a string? I've just been using `NSJSONSerialization.dataWithJSONObject:error:` to convert my `AnyObject` to the appropriate `NSData` then attach it to the `NSRequest`. You don't need to maintain type info or worry about converting it to a string manually.

nvh:
[...]

PR #47, opened 2015-02-13
Feature/json encodable

nvh:
As discussed in #42 this is an implementation of JSONEncodable.

A question I encountered is what should happen to optional values when represented in a JSON object. It's possible to include the key, but and make the value `null` or just remove the key entirely. I chose to include the keys and make it easy to filter them afterwards.

Let me know what you think!

### DIFF
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		B351CE721A80ACE8008E20D0 /* JSONEncodable.swift in Sources */ = {isa = PBXBuildFile; fileRef = B351CE711A80ACE8008E20D0 /* JSONEncodable.swift */; };
 		EA08313119D5EEAF003B90D7 /* TypeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA08313019D5EEAF003B90D7 /* TypeTests.swift */; };
 		EA08313319D5EEF2003B90D7 /* post_comments.json in Resources */ = {isa = PBXBuildFile; fileRef = EA08313219D5EEF2003B90D7 /* post_comments.json */; };
 		EA08313519D5EFC0003B90D7 /* types.json in Resources */ = {isa = PBXBuildFile; fileRef = EA08313419D5EFC0003B90D7 /* types.json */; };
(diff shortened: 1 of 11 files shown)


gfontenot:
I'm still honestly not sure why this is useful. It seems like this is a lot of code (and boilerplate) to add for creating dictionary representations of your models. We haven't seen the need to move from a model back to `JSONValue`. Where are you running into this?

nvh:
When sending objects back to the server, I want to be sure that the data that I provide to the JSON serialisation, is actually of the correct type at compile time.

gfontenot:
That's good in theory, but you aren't gaining that here. There's no difference at compile time between the types `JSONValue.JSONString("foo")` and `JSONValue.JSONNumber(2)`. In order to get some kind of compile-time check on your JSON, you'd need to have a stronger type system than Swift allows.

gfontenot:
@nvh Did you have any other thoughts on this?

nvh:
True, but there is a difference between trying to encode

```swift
struct Thing {
  let date: NSDate
  let color: UIColor 
  func encode() -> AnyObject {
     ["date":date,"color":color]
  }
}
```
This will compile and run fine, and probably turn into json too, but you have no control on how it will do so, while if the protocol is

```swift
  func encode() -> JSONValue {
     ["date":date,"color":color]
  }
```
The compiler will complain that NSDate and UIColor are not JSONEncodable and you will have to specify explicitly how to convert them to something that can be JSON encoded.

tonyd256:
I get what you're saying but that seems like the only value we'd be getting from adding this. I have a couple thoughts.

I don't think the benefits outweigh the bloat from the code additions and the overhead for a user to implement this. If we could use this to guarantee type safety of the resulting JSON then I think it would be awesome, but because of JSON's dynamic nature and the fact that the API backend is what controls the types needed, I think this would be better suited for tests in the UnitTest bundle.

Also, when I'm thinking of the scope of Argo, I see it as a framework for translating JSON into model objects. This is the core of Argo and the reason for it being built. More over than type safety, it is non trivial to parse JSON (dynamically typed) into Swift (static typed) model so that's the biggest benefit it can provide. Going the other way around is easier to put together without Argo. I don't think Swift can provide the type safety we'd like if we were to add this.

However, I do think this is an interesting problem and could be better suited as a separate framework that either used Argo as a dependency or came up with a better version of `JSONValue` that we could share.

Thanks for putting in the work, but I'm going to close this for now because it's out of scope for Argo.

vesterbaek:
@nvh I think your contributions here are useful. Have you considered making an "Argo-encode" library for this?

nvh:
@vesterbaek Yes, that was actually my next step, but I havent gotten to it yet

PR #48, opened 2015-02-14
Floats not JSONDecodable

kaishin:
```swift
import Foundation
import Argo
import Runes

struct Location {
  let latitude: Float
  let longitude: Float
}

extension Location: JSONDecodable {
  static func create(latitude: Float)(longitude: Float) -> Location {
    return Location(latitude: latitude, longitude: longitude)
  }

  static func decode(json: JSONValue) -> Location? {
    return Location.create
      <^> json <| "latitude"
      <*> json <| "longitude"
  }
}
```

This structs causes SourceKit to hang forever. I got no error at first, until I removed all other properties, then I got this: `Type Float does not conform to protocol JSONDecodable`.

### DIFF
--- a/Argo/Globals/JSON.swift
+++ b/Argo/Globals/JSON.swift
@@ -23,3 +23,9 @@ extension Bool: JSONDecodable {
     return j.value()
   }
 }
+
+extension Float: JSONDecodable {
+  public static func decode(j: JSONValue) -> Float? {
+    return j.value()
+  }
+}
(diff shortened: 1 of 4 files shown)


Argo/Globals/JSON.swift
```diff
@@ -23,3 +23,9 @@ extension Bool: JSONDecodable {
     return j.value()
   }
 }
+
+extension Float: JSONDecodable {
+  public static func decode(j: JSONValue) -> Bool? {
```
tonyd256: This should probably be returning a `Float?`

tonyd256: I'm surprised this even compiled

tonyd256:
:+1: squash and merge my man

PR #50, opened 2015-02-16
Use Carthage as an actual development dependency

gfontenot:
This moves us away from Git Submodules and towards using Carthage
directly as a development dependency. Using the carthage-built
frameworks directly means we can also use cool things like Carthage's
support for pre-built binaries.


tonyd256:
looks good to me :+1:

PR #51, opened 2015-02-16
Clarify that Runes isn't an optional dependency

gfontenot:
Runes needs to be added to the client application in order for Argo to work.
Given that it isn't an optional dependency, we should be clearer about that in
the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,9 @@ The Greek word for _swift_ and the ship used by Jason, son of Aeson, of the
 Argonauts. Aeson is the JSON parsing library in Haskell that inspired Argo,
 much like Aeson inspired his son Jason.
 
-NOTE: The master branch of Argo is pushing ahead with Swift 1.2 support. Support for 
-Swift 1.1 can be found on branch [swift-1.1] and with tag version v0.3.2.
+NOTE: The master branch of Argo is pushing ahead with Swift 1.2 support.
+Support for Swift 1.1 can be found on branch [swift-1.1] and in the 0.3.x
+versions of tags/releases.
 
 [swift-1.1]: https://github.com/thoughtbot/Argo/tree/swift-1.1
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,8 +5,9 @@ The Greek word for _swift_ and the ship used by Jason, son of Aeson, of the
 Argonauts. Aeson is the JSON parsing library in Haskell that inspired Argo,
 much like Aeson inspired his son Jason.
 
-NOTE: The master branch of Argo is pushing ahead with Swift 1.2 support. Support for 
-Swift 1.1 can be found on branch [swift-1.1] and with tag version v0.3.2.
+NOTE: The master branch of Argo is pushing ahead with Swift 1.2 support.
+Support for Swift 1.1 can be found on branch [swift-1.1] and with tag version
+v0.3.2.
```
tonyd256: how about changing this to 0.3.x since it will be changing

tonyd256:
one comment then :+1:

README.md
gfontenot: Good call.

gfontenot:
@tonyd256 Want to take another look at the new language for the tag version?

README.md
```diff
@@ -5,8 +5,9 @@ The Greek word for _swift_ and the ship used by Jason, son of Aeson, of the
 Argonauts. Aeson is the JSON parsing library in Haskell that inspired Argo,
 much like Aeson inspired his son Jason.
 
-NOTE: The master branch of Argo is pushing ahead with Swift 1.2 support. Support for 
-Swift 1.1 can be found on branch [swift-1.1] and with tag version v0.3.2.
+NOTE: The master branch of Argo is pushing ahead with Swift 1.2 support.
+Support for Swift 1.1 can be found on branch [swift-1.1] and in the 0.3.x
+branch of tags/releases.
```
tonyd256: how about `and in the 0.3.x versions or tags/releases` ... they aren't really branches right?

tonyd256:
:+1:

PR #52, opened 2015-02-17
Pin Runes to version 1.1.1 for 0.3.x

gfontenot:
Since we weren't pinning the version of Runes with Carthage, users were
getting the version of Runes compatible with Swift 1.2, which obviously
isn't what we wanted. Pinning this to 1.1.1 will ensure that doesn't
happen in the future.

This also makes the same changes to the project that we did for master,
removing the submodules and using the binaries provided by Carthage.


felixjendrusch:
That was quick! :+1:

tonyd256:
:+1:

Keithbsmiley:
:+1:

felixjendrusch:
Just for reference, [here](https://github.com/Carthage/Carthage/issues/339) is the original issue.

PR #55, opened 2015-02-18
Fix EXC_BAD_ACCESS on 32 bit systems

### LINKED ISSUE #54: EXC_BAD_ACCESS when running tests on 32-bit simulator
I'm using the Xcode 6.3-Beta and the latest Argo from master. When I'm running tests for example on iPhone 5 simulator it crashes on first attempt to map json property to model property. For example in

EmbeddedJSONDecodingTests.swift:
```
func testCommentDecodingWithEmbeddedUserName() {
    let json = JSONValue.parse <^> JSONFileReader.JSON(fromFile: "comment")
    self.json2 = json
    let comment = self.json2 >>- Comment.decode

    XCTAssert(comment != nil)
    XCTAssert(comment?.id == "6")
    XCTAssert(comment?.text == "Cool story bro.")
    XCTAssert(comment?.authorName == "Cool User")
  }
```
Comment.swift:
```
return Comment.create
      <^> j <| "id" // Crashes on this line
      <*> j <| "text"
      <*> j <| ["author", "name"]
```

This happens also when running app which uses Argo on the simulator. When running tests on iPhone 6 simulator it does not crash.

gfontenot:
It looks like even in Swift 1.2, partially applying static functions on a
generic protocol object causes a crash, but only on 32 bit systems. We had
originally thought these were fixed with Swift 1.2, but apparently the fix
didn't completely take for 32 bit systems.

Fixes #54

### DIFF
--- a/Argo/Globals/JSONValue.swift
+++ b/Argo/Globals/JSONValue.swift
@@ -31,7 +31,7 @@ public extension JSONValue {
 
   static func mapDecode<A where A: JSONDecodable, A == A.DecodedType>(value: JSONValue) -> [A]? {
     switch value {
-    case let .JSONArray(a): return sequence(A.decode <^> a)
+    case let .JSONArray(a): return sequence({ A.decode($0) } <^> a)
     default: return .None
     }
   }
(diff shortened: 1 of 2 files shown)


tonyd256:
:middle_finger: :32_bit:

PR #56, opened 2015-02-18
Update readme parse function to not return optional

Keithbsmiley:
Looks like this got a little out of date


gfontenot:
:sparkles: :bomb: :tada:

PR #57, opened 2015-02-20
Remove value function

tonyd256:
The value function is not needed because we can pull the values out of the
JSONValue enum with a switch statement. This was also the source of some
confusion with `JSONValue.value()` and unnecessary complexity with more
type inference.


gfontenot:
ping @thoughtbot/ios @jferris

gfontenot:
I dig it.

kaishin:
:+1:

jnutting:
+1

jferris:
Looks good to me.

PR #58, opened 2015-02-20
Move JSONValue.mapDecode to a global function

gfontenot:
This was kind of bolted onto JSONValue without too much benefit. It was
also pretty poorly named. It makes more sense to me as a global
function, seeing as it's essentially already a workaround for the fact
that we can't create an extension on Array with a type constraint of
JSONDecodable.

### DIFF
--- a/Argo/Globals/JSON.swift
+++ b/Argo/Globals/JSON.swift
@@ -1,4 +1,5 @@
 import Foundation
+import Runes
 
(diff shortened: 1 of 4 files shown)


tonyd256:
ping  @thoughtbot/ios

kaishin:
:+1:

jnutting:
+1

PR #59, opened 2015-02-20
Rename objects to remove JSON prefix

gfontenot:
We don't really need to use JSON as a prefix everywhere.

When this enum was initially written, we thought it would be an internal
implementation detail that wasn't ever exposed to users. But as we've
been using it (and helping others use it) in the field, exposing this
publicly has been immensely powerful. Given that this is the case, we
should really clean up the naming.

Weird things to note:

1. Needing to use `Swift.String` inside the `JSON` enum to help the
compiler know which string we're referring to (no clue why we don't need
this for `Array`)

2. Renaming the file containing the default decoding functions for types
in the standard lib to StandardTypes. There must be a better name for
this, but I can't come up with it.

3. Renaming `JSONDecodable` to `Decodable`. I don't know if this is
needed, or if it's too vague. This is the one place that I feel OK
keeping `JSON` around.


tonyd256:
Looks good to me but I would prefer the protocol remain `JSONDecodable`. Just `Decodable` is too vague and actually with the renaming on `JSONValue` to `JSON`, I think `JSONDecodable` is a much nicer name.

tonyd256:
ping  @thoughtbot/ios

kaishin:
+1 for Tony's suggestion. `Decodable` is not precise enough.

jnutting:
I think that Decodable is a bit vague, yeah, but OTOH JSONDecodable is a bit too *precise*. Because when you think about it, Argo isn't really about JSON exactly, it's is really about constructing things from the contents out of dictionaries, arrays, etc... the same sort of thing that you can have in JSON, but also in property lists (You can even think of JSON as a transport format for property lists). So if it's called JSONDecodable, it could just as well be called PropertyListDecodable! Which is not what I'm really suggesting. But maybe a name that applies to the structure we're pulling from: DictionaryDecodable, or KeyValueDecodable, or... ArgoDecodable?

gfontenot:
@jnutting Fantastic point. I've thought about that as well. In theory, as long as users are creating `[String: AnyObject]` dictionaries, they _should_ be able to decode them with Argo. Someone building a YAML parser, for example, could use Argo to transform that into a dictionary.

The only restriction here is that (as Argo works now) the parser would need to parse any numeric value into an `NSNumber`. That includes `Bool` values.

I think for now, we can focus on JSON parsing, but mainly because I don't have any good ideas for naming. I've changed the protocol name back to `JSONDecodable`

gfontenot:
Another thought, related to that last one: Aeson uses `Value` as the name of their enum. Should we follow their lead? They then use `fromJSON` as their version of `decode` though. /cc @jferris

jferris:
I think `Value` is a fine name as long as you can namespace it. I'm not familiar enough with Swift's module system to know how you could handle a situation where you need to work with two names like `Value` (maybe you have a JSON parser and a YAML parser or something).

gfontenot:
@jferris Namespacing comes from framework names. So this version of `Value` would be `Argo.Value` if it's ambiguous which we mean.

gfontenot:
#64 is related to this PR

jferris:
That seems cool to me.

If you want to follow Aeson, `Value` makes sense and is easy to understand when you know the context is JSON parsing.

If you want it to be easy to guess what's happening without knowing Argo beforehand, using something like 'JSON' might be more obvious.

gfontenot:
I think we should probably stick to `JSON` for now. `Value` is fairly vague and easily clobbered. Especially during this period of education around these concepts, helping users along with (maybe overly) specific types seems like the best course of action to me.

tonyd256:
:+1:

kaishin:
:+1:

PR #65, opened 2015-02-24
Argo is bigger than JSON

tonyd256:
With Argo we can parse anything that is a dictionary or array into model
objects. This add tests for Swift Dictionaries and plists.

### DIFF
--- a/ArgoTests/Tests/PListDecodingTests.swift
+++ b/ArgoTests/Tests/PListDecodingTests.swift
@@ -1,0 +1,25 @@
+import XCTest
+import Argo
+import Runes
+
+class PListDecodingTests: XCTestCase {
+  func testDecodingAllTypesFromPList() {
+    let value = JSON.parse <^> PListFileReader.plist(fromFile: "types")
+    let model = value >>- TestModel.decode
+
+    XCTAssert(model != nil)
+    XCTAssert(model?.int == 5)
+    XCTAssert(model?.string == "Cooler User")
+    XCTAssert(model?.double == 3.4)
+    XCTAssert(model?.float == 1.1)
+    XCTAssert(model?.bool == false)
+    XCTAssert(model?.intOpt == nil)
+    XCTAssert(model?.stringArray.count == 2)
+    XCTAssert(model?.stringArrayOpt == nil)
+    XCTAssert(model?.eStringArray.count == 2)
+    XCTAssert(model?.eStringArrayOpt != nil)
+    XCTAssert(model?.eStringArrayOpt?.count == 0)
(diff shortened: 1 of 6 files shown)


kaishin:
Do we want to close an eye on the `JSON.parse` for now?

gfontenot:
@kaishin What do you mean?

ArgoTests/Tests/PListDecodingTests.swift
```diff
@@ -0,0 +1,25 @@
+import XCTest
+import Argo
+import Runes
+
+class PListDecodingTests: XCTestCase {
+  func testDecodingAllTypesPList() {
```
gfontenot: maybe `testDecodingAllTypesFromPList`?

ArgoTests/Tests/SwiftDictionaryDecodingTests.swift
```diff
@@ -0,0 +1,55 @@
...
+      "email": "some@tester.com"
+    ]
+    let value = JSON.parse(userDict)
+    let user = User.decode(value)
+    XCTAssert(user != nil)
+    XCTAssert(user?.id == 5)
+    XCTAssert(user?.name == "Tester")
+  }
+
+  func testDecodingSwiftAllTypesDictionary() {
```
gfontenot: Maybe add `From` in this test name as well

ArgoTests/Tests/SwiftDictionaryDecodingTests.swift
```diff
@@ -0,0 +1,55 @@
+import XCTest
+import Argo
+
+class SwiftDictionaryDecodingTests: XCTestCase {
+  func testDecodingSwiftUserDictionary() {
```
gfontenot: What does this test provide that the next one doesn't?

gfontenot:
couple small naming nitpicks but it's nice to know that this works as expected.

ArgoTests/Tests/SwiftDictionaryDecodingTests.swift
tonyd256: nothing ... it was a first try that I left in there ... it can be removed

kaishin:
@gfontenot Not sure if I got this right, but it doesn't seem like we're parsing JSON here.

gfontenot:
@kaishin Sorry, you're right, and I get that. I wasn't sure what you were proposing. Changing the name of `JSON`?

tonyd256:
I think that's a fine proposition but was out of scope for this PR. I agree thought ... let's think about that

PR #66, opened 2015-02-24
Remove JSON subscripting

### LINKED ISSUE #63: Have JSONValue conform to JSONDecodable
If `JSONValue` conforms to `JSONDecodable`, we can cut the number of operators we define in half in favor of using chained parsing operators.

`j <| ["foo", "bar"]` would become `j <| "foo" <| "bar"`

This would also allow us to get rid of `subscript`, as well as `find` probably.

There might be performance implications here.

gfontenot:
This removes the ability to subscript JSON values, instead opting to
centralize around the `<|` operators for pulling out values.

Fixes #63

### DIFF
--- a/Argo/Globals/fold.swift
+++ b/Argo/Globals/fold.swift
@@ -1,7 +1,0 @@
-import Runes
-
-func foldM<S: SequenceType, U>(sequence: S, initial: U, combine: (U, S.Generator.Element) -> U?) -> U? {
-  return reduce(sequence, initial) { accum, x in
-    accum >>- { combine($0, x) }
-  }
-}
--- a/Argo/Globals/fold.swift
+++ b/Argo/Globals/fold.swift
@@ -1,0 +1,7 @@
+import Runes
+
+func foldM<S: SequenceType, U>(sequence: S, initial: U, combine: (U, S.Generator.Element) -> U?) -> U? {
+  return reduce(sequence, initial) { accum, x in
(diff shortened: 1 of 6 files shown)


tonyd256:
:+1:

gfontenot:
ping @thoughtbot/ios how do y'all feel about this now?

Argo/Globals/fold.swift
```diff
@@ -0,0 +1,7 @@
+import Runes
+
+func foldM<S: SequenceType, U>(sequence: S, initial: U, combine: (U, S.Generator.Element) -> U?) -> U? {
```
Keithbsmiley: Not sure where you ever landed on using names from haskell, but this name makes no sense otherwise.

Keithbsmiley:
Definitely better syntax!

Argo/Globals/fold.swift
gfontenot: This is completely internal, so I'm not super duper worried about that, honestly.

Keithbsmiley: Ok


--
Keith Smiley

jferris: If you're looking for a more Swifty name, maybe something like `flatReduce`?

gfontenot: I can get behind that.

kaishin:
As I mentioned before, I am totally onboard for this.

Issue #67, opened 2015-02-24
Return `Parser<T>` instead of `Optional<T>`


gfontenot:
Aeson returns a `Parser` type, not a `Maybe` type. Generalizing away from `Optional` might help us support other return formats in the future (like `Result`)

tonyd256:
addressed in #72

PR #72, opened 2015-02-26
Implement a Parser type for error messaging

tonyd256:
Using a Parser type allows us to abstract the return type away from the decode
function. This shifts some naming to what Aeson uses for an easier comparison
to their technique. The old `decode` function is now `fromJSON` and the new
`decode` function is a global function with multiple implementations for
different return types. It also abstracts away the JSON.parse function. The
user can now just call decode on their JSON and get back the Optional decoded
type, or if they prefer they can get the Parser type directly and use it to
view errors.

With Parser, we also don't have dependence on the return type. Someone using a
Result framework like LlamaKit could implement a global decode function for
Result and everything internal would work as normal.

### DIFF
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -18,6 +18,10 @@
 		EA395DCA1A52FC1400EB607E /* root_object.json in Resources */ = {isa = PBXBuildFile; fileRef = EA395DC91A52FC1400EB607E /* root_object.json */; };
 		EA395DCB1A52FC1400EB607E /* root_object.json in Resources */ = {isa = PBXBuildFile; fileRef = EA395DC91A52FC1400EB607E /* root_object.json */; };
 		EA4EAF7319DD96330036AE0D /* types_fail_embedded.json in Resources */ = {isa = PBXBuildFile; fileRef = EA4EAF7219DD96330036AE0D /* types_fail_embedded.json */; };
+		EA8148B61A9E476D00683AFF /* Decoded.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA8148B51A9E476D00683AFF /* Decoded.swift */; };
+		EA8148B71A9E476D00683AFF /* Decoded.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA8148B51A9E476D00683AFF /* Decoded.swift */; };
+		EA8148B91A9E573300683AFF /* decode.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA8148B81A9E573300683AFF /* decode.swift */; };
+		EA8148BA1A9E573300683AFF /* decode.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA8148B81A9E573300683AFF /* decode.swift */; };
 		EABDF68A1A9CD46100B6CC83 /* SwiftDictionaryDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EABDF6891A9CD46100B6CC83 /* SwiftDictionaryDecodingTests.swift */; };
 		EABDF68B1A9CD46400B6CC83 /* SwiftDictionaryDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EABDF6891A9CD46100B6CC83 /* SwiftDictionaryDecodingTests.swift */; };
 		EABDF68F1A9CD4EA00B6CC83 /* PListFileReader.swift in Sources */ = {isa = PBXBuildFile; fileRef = EABDF68D1A9CD4EA00B6CC83 /* PListFileReader.swift */; };
@@ -26,6 +30,10 @@
 		EABDF6921A9CD4EA00B6CC83 /* types.plist in Resources */ = {isa = PBXBuildFile; fileRef = EABDF68E1A9CD4EA00B6CC83 /* types.plist */; };
 		EABDF6941A9CD4FC00B6CC83 /* PListDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EABDF6931A9CD4FC00B6CC83 /* PListDecodingTests.swift */; };
 		EABDF6951A9CD4FC00B6CC83 /* PListDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EABDF6931A9CD4FC00B6CC83 /* PListDecodingTests.swift */; };
+		EAC59A0F1A9F74C500BEB5CC /* DecodedOperators.swift in Sources */ = {isa = PBXBuildFile; fileRef = EAC59A0E1A9F74C500BEB5CC /* DecodedOperators.swift */; };
+		EAC59A101A9F74C500BEB5CC /* DecodedOperators.swift in Sources */ = {isa = PBXBuildFile; fileRef = EAC59A0E1A9F74C500BEB5CC /* DecodedOperators.swift */; };
+		EAD8BF621AA0DD3900A11963 /* Box.framework in CopyFiles */ = {isa = PBXBuildFile; fileRef = EAEE2F4B1A9FCC9F00DA2846 /* Box.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		EAD8BF631AA0DD4100A11963 /* Box.framework in CopyFiles */ = {isa = PBXBuildFile; fileRef = EAEE2F4D1A9FCCAA00DA2846 /* Box.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		EAD9FAEB19D0F6930031E006 /* Operators.swift in Sources */ = {isa = PBXBuildFile; fileRef = EAD9FAEA19D0F6930031E006 /* Operators.swift */; };
 		EAD9FAF419D0F74C0031E006 /* StandardTypes.swift in Sources */ = {isa = PBXBuildFile; fileRef = EAD9FAF319D0F74C0031E006 /* StandardTypes.swift */; };
 		EAD9FAF619D0F7900031E006 /* JSONOperators.swift in Sources */ = {isa = PBXBuildFile; fileRef = EAD9FAF519D0F7900031E006 /* JSONOperators.swift */; };
@@ -45,6 +53,8 @@
 		EAD9FB1A19D4B23F0031E006 /* JSON.swift in Sources */ = {isa = PBXBuildFile; fileRef = EAD9FB1919D4B23F0031E006 /* JSON.swift */; };
 		EADADCB21A5DB6F600B180EC /* EquatableTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EADADCB11A5DB6F600B180EC /* EquatableTests.swift */; };
 		EADADCB41A5DB7F800B180EC /* EquatableTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EADADCB11A5DB6F600B180EC /* EquatableTests.swift */; };
+		EAEE2F4C1A9FCC9F00DA2846 /* Box.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = EAEE2F4B1A9FCC9F00DA2846 /* Box.framework */; };
+		EAEE2F4E1A9FCCAA00DA2846 /* Box.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = EAEE2F4D1A9FCCAA00DA2846 /* Box.framework */; };
 		F802D4C31A5EE061005E236C /* NSURL.swift in Sources */ = {isa = PBXBuildFile; fileRef = F802D4C21A5EE061005E236C /* NSURL.swift */; };
 		F802D4C41A5EE172005E236C /* NSURL.swift in Sources */ = {isa = PBXBuildFile; fileRef = F802D4C21A5EE061005E236C /* NSURL.swift */; };
 		F802D4C61A5EE2D5005E236C /* url.json in Resources */ = {isa = PBXBuildFile; fileRef = F802D4C51A5EE2D5005E236C /* url.json */; };
@@ -118,6 +128,7 @@
 			dstSubfolderSpec = 10;
 			files = (
 				F87D9D8D1A92676F00C8AF1D /* Runes.framework in CopyFiles */,
+				EAD8BF621AA0DD3900A11963 /* Box.framework in CopyFiles */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -128,6 +139,7 @@
 			dstSubfolderSpec = 10;
 			files = (
 				F87D9D8F1A92677A00C8AF1D /* Runes.framework in CopyFiles */,
+				EAD8BF631AA0DD4100A11963 /* Box.framework in CopyFiles */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -159,10 +171,13 @@
 		EA395DC61A52F93B00EB607E /* ExampleTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = ExampleTests.swift; sourceTree = "<group>"; };
 		EA395DC91A52FC1400EB607E /* root_object.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = root_object.json; sourceTree = "<group>"; };
 		EA4EAF7219DD96330036AE0D /* types_fail_embedded.json */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.json; path = types_fail_embedded.json; sourceTree = "<group>"; };
+		EA8148B51A9E476D00683AFF /* Decoded.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = Decoded.swift; sourceTree = "<group>"; };
+		EA8148B81A9E573300683AFF /* decode.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = decode.swift; sourceTree = "<group>"; };
 		EABDF6891A9CD46100B6CC83 /* SwiftDictionaryDecodingTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = SwiftDictionaryDecodingTests.swift; sourceTree = "<group>"; };
 		EABDF68D1A9CD4EA00B6CC83 /* PListFileReader.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = PListFileReader.swift; sourceTree = "<group>"; };
 		EABDF68E1A9CD4EA00B6CC83 /* types.plist */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = text.plist.xml; path = types.plist; sourceTree = "<group>"; };
 		EABDF6931A9CD4FC00B6CC83 /* PListDecodingTests.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = PListDecodingTests.swift; sourceTree = "<group>"; };
+		EAC59A0E1A9F74C500BEB5CC /* DecodedOperators.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = DecodedOperators.swift; sourceTree = "<group>"; };
 		EAD9FACF19D0EAB50031E006 /* Argo.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; includeInIndex = 0; path = Argo.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		EAD9FAD319D0EAB50031E006 /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
 		EAD9FADA19D0EAB60031E006 /* ArgoTests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = ArgoTests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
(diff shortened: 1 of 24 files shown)


Argo/Globals/Parser.swift
```diff
@@ -0,0 +1,29 @@
...
+
+  public var value: T? {
+    switch self {
+    case let .Success(box): return box.value
+    default: return .None
+    }
+  }
+}
+
+public func typeMismatch<T>(type: String, object: JSON) -> Parser<T> {
```
gfontenot: should these be `public`?

Argo/Globals/Parser.swift
```diff
@@ -0,0 +1,29 @@
...
+
+public func parseOptional<A>(p: Parser<A>) -> Parser<A?> {
+  switch p {
+  case let .Success(box): return .Success(Box(.Some(box.value)))
+  case let .MissingKey(string): return .Success(Box(.None))
+  case let .TypeMismatch(string): return .TypeMismatch(string)
+  }
+}
+
+public final class Box<T> {
```
gfontenot: Should this be `public`? If so, should we rely on [robrix/Box](https://github.com/robrix/Box)?

tonyd256: yup I think it would be fine to use an external library here ... it has to be public because Parse it.

Argo/Globals/Parser.swift
jferris: Should this be a private function in `StandardTypes.swift`?

Argo/Globals/decode.swift
```diff
@@ -0,0 +1,15 @@
+public func decode<T: JSONDecodable where T == T.DecodedType>(object: AnyObject) -> T? {
+  return decodeResult(object).value
+}
+
+public func decode<T: JSONDecodable where T == T.DecodedType>(object: AnyObject) -> [T]? {
+  return decodeResult(object).value
+}
+
+public func decodeResult<T: JSONDecodable where T == T.DecodedType>(object: AnyObject) -> Parser<T> {
```
gfontenot: Couldn't this just be named `decode` as well? If not, it should be `decodeParser` instead of `decodeResult`.

Argo/Globals/Parser.swift
jferris: Does this need to be `public`? If you can avoid exposing the `Box` type, that seems better.

Argo/Globals/Parser.swift
tonyd256: nope they don't have to be ... just convenience for the user if anything

Argo/Globals/decode.swift
tonyd256: yup you're right ... it can just be `decode`

Argo/Globals/maybe.swift
```diff
@@ -0,0 +1,6 @@
+public func maybe<A, B>(b: B, f: A -> B, a: A?) -> B {
```
gfontenot: `maybe` seems like a weird name here. Should this be called `optional`? Also, the type signature feels un-swifty to me.

Also, why isn't this just `f <^> a ?? b`?

Argo/Globals/Parser.swift
gfontenot: @tonyd256 No, exposing `Parser` doesn't expose `Box`. This should really be `internal`. We'd only need to expose this if we allowed users to get at the boxed value.

Argo/Globals/Parser.swift
gfontenot: I don't think the user needs access to this. We should probably default to `internal`, instead of defaulting to `public`.

Argo/Globals/Parser.swift
```diff
@@ -0,0 +1,29 @@
...
+    default: return .None
+    }
+  }
+}
+
+public func typeMismatch<T>(type: String, object: JSON) -> Parser<T> {
+  return .TypeMismatch("\(object) is not a \(type)")
+}
+
+public func parseOptional<A>(p: Parser<A>) -> Parser<A?> {
```
gfontenot: This feels poorly named to me.

Argo/Globals/maybe.swift
tonyd256: hmm ya i guess that's all it is ... weird how that appeared after I broke down the types.

gfontenot: I really (really) dislike the introduction of this function, so if we can get rid of it, that would be great.

Argo/Globals/Parser.swift
gfontenot: So it looks like we _would_ need to expose this to users in order to let them write custom parsers. In that case, can we move these onto `Parser` as static functions? We can rely on type inference to let users write `.typeMismatch` instead of `Parser.typeMismatch`, and we won't pollute the global scope.

Argo/Globals/maybe.swift
tonyd256: it's already gone

Argo/Globals/Parser.swift
tonyd256: I agree ... I have no clue what to call this

Argo/Globals/Parser.swift
tonyd256: @gfontenot If I make Parser public and it's in a case on Parser then Box has to be public ... really, not kidding ... I even double checked

gfontenot: Oh, right. Gross. Well, we should at least rely on an existing implementation then.

Argo/Globals/Parser.swift
tonyd256: ya I made that function private and users can just user the type inferred cases for now

Argo/Globals/Parser.swift
tonyd256: agreed

Argo/Globals/Parser.swift
gfontenot: But we'd need this to be `public` in order to let users write more complex parsers. Like, for example, the ability to write parsers for `NSURL` (like we did in the tests).

Argo/Globals/Parser.swift
```diff
@@ -0,0 +1,22 @@
...
+
+  public var value: T? {
+    switch self {
+    case let .Success(box): return box.value
+    default: return .None
+    }
+  }
+}
+
+func parseOptional<A>(p: Parser<A>) -> Parser<A?> {
```
gfontenot: Just making a note here that we should figure out a better name.

Argo/Globals/StandardTypes.swift
```diff
@@ -2,53 +2,65 @@ import Foundation
...
+public func decodeArray<A where A: JSONDecodable, A == A.DecodedType>(value: JSON) -> Parser<[A]> {
   switch value {
-  case let .Array(a): return sequence({ A.decode($0) } <^> a)
-  default: return .None
+  case let .Array(a): return sequence({ A.fromJSON($0) } <^> a)
+  default: return typeMismatch("Array", value)
   }
 }
+
+public func decodeObject(value: JSON) -> Parser<Dictionary<String, JSON>> {
```
gfontenot: Why do we all of a sudden need this function?

tonyd256: it was to help pull out a dictionary from a JSON so the case to use the maybe pattern became obvious

gfontenot: It's odd to me that this acts differently than every other `decode*` function. Everything else returns a fully decoded object, while this still returns `JSON`. I'm not sure I understand why we moved to this from doing this inside the `<|` operator like we did originally.

Argo/Globals/StandardTypes.swift
```diff
@@ -2,53 +2,65 @@ import Foundation
...
+  switch value {
+  case let .Object(o): return pure(o)
+  default: return typeMismatch("Object", value)
+  }
+}
+
+private func typeMismatch<T>(type: String, object: JSON) -> Parser<T> {
```
gfontenot: I do think this is useful enough for users that want to write their own parsers that it should probably be `public`. But I still think it should be a static method on `Parser` instead of a free function.

gfontenot: I'd also rename `type` to `expectedType` for clarity.

Argo/Globals/StandardTypes.swift
jferris: The better way would be to have `Dictionary` implement `JSONDecodable`, but we can't make protocol extensions for abstract types.

This followed the similar solution used for `Array`, and simplified the `<|` function, which was becoming somewhat complex. It also consolidates the type-checking in one file/module.

jferris: Talking about this a little in person, it probably makes sense to use `Parser<Dictionary<String, A>>` where `A` is `JSONDecoable`

gfontenot: We should also use the shorthand for the dictionary type.

gfontenot:
Fixes #67

Argo/Globals/StandardTypes.swift
jferris: I don't think this needs to be exposed. Argo covers all of the primitive types in JSON, so we should be able to always provide a good error message like "Expected post.user.name to be a String." Overriding it to be "Expected post.user to be a User" would actually be less useful; I think you'll always want to know the primitive type you were expecting.

Argo/Globals/StandardTypes.swift
tonyd256: Can someone explain to me how it makes sense now to have this be A ... more than the argument "because the other are"

jferris: I believe that if you:

* Change the requirement from `JSON` to `A` where `A` is `JSONDecodable`
* Change `pure` to `A.fromJSON`

Then you'll be able to ask for other dictionaries. For example, if you want a dictionary from `String` to `Boolean`, it will Just Work. It would also work for `String` to `User` if you wanted that, or any other case where the dictionary value is `JSONDecodable`.

Argo/Globals/StandardTypes.swift
```diff
@@ -2,53 +2,60 @@ import Foundation
...
+public func decodeArray<A where A: JSONDecodable, A == A.DecodedType>(value: JSON) -> Parser<[A]> {
   switch value {
-  case let .Array(a): return sequence({ A.decode($0) } <^> a)
-  default: return .None
+  case let .Array(a): return sequence({ A.fromJSON($0) } <^> a)
+  default: return .typeMismatch("Array", object: value)
+  }
+}
+
+public func decodeObject(value: JSON) -> Parser<[String:JSON]> {
```
gfontenot: Space after `:`

Argo/Globals/StandardTypes.swift
tonyd256: The return type of this function is currently `Parser<[String: JSON]>` ... we use a switch to pull out the object in the form of `[String: JSON]` and then apply `pure` to get the parser. If we change JSON to A and then our return type looks like this `Parser<[String: A]>` then when we pull out the `[String: JSON]` you want to apply `A.fromJSON` to it but that won't work because `fromJSON takes a JSON not a [String:JSON] ...

Argo/Globals/StandardTypes.swift
gfontenot: Good point. So if users always parse into the primitive types, then pass that into their own parsers (like a function that vends `NSURL`s), they won't need this at all.

Argo/Globals/StandardTypes.swift
tonyd256: even if we didn't pull out the `[String: JSON]` from the input and just passed the input into `A.fromJSON` it would return a `Parser<A>` not a `Parser<[String:A]>`

Argo/Globals/StandardTypes.swift
tonyd256: ok so make it private again?

Argo/Globals/StandardTypes.swift
gfontenot: Right. We'd need to `map` `A.fromJSON` over the values in the object, which would leave us with `[String: Parser<A>]`, then we could use `sequence` to get to `Parser<[String: A]>`. We would need to write both `map` and `sequence` for Dictionaries in terms of the values.

Argo/Globals/StandardTypes.swift
gfontenot: Yep. Sorry about that. This was probably best constrained to `StandardTypes.swift`

Argo/Globals/Parser.swift
```diff
@@ -0,0 +1,26 @@
...
+    default: return .None
+    }
+  }
+
+  public static func typeMismatch(expectedType: String, object: JSON) -> Parser<T> {
+    return .TypeMismatch("\(object) is not a \(expectedType)")
+  }
+}
+
+func parseOptional<A>(p: Parser<A>) -> Parser<A?> {
```
gfontenot: Bringing back the conversation about this sucker:

Is it possible that this is doing the wrong thing, and that's why it's hard to name? Is it possible that `.TypeMismatch` should _also_ be coerced into `.Success(.None)`? If that were the case, it seems like it would simplify this function conceptually, and also seems like it might end up with a clearer name?

tonyd256: we decided that a "key missing" is an acceptable error to ignore when the user uses the `<|?` operator because they are saying "This may or may not be there"; however, if it is there and it fails from casting we want that error to carry through because that's something we think the user should address.

jferris: Right. Picture that the response contains an optional email address, and then the server switches to sending an array of email addresses.

If you fail on type mismatch, you'll catch that immediately.

If you ignore type mismatch, your parser will claim that it's cool every time and you'll wonder why the server isn't sending any usernames.

gfontenot: So how about dropping `parse` from the name, making it a static function on `Parser` and just calling it `optional`. Does that make sense?

tonyd256: it was originally `optional` but i thought the name was too vague ... not that `parseOptional` is better ... I'm find with that though

Argo/Globals/StandardTypes.swift
```diff
@@ -2,53 +2,60 @@ import Foundation
...
-  case let .Array(a): return sequence({ A.decode($0) } <^> a)
-  default: return .None
+  case let .Array(a): return sequence({ A.fromJSON($0) } <^> a)
+  default: return .typeMismatch("Array", object: value)
+  }
+}
+
+public func decodeObject<A where A: JSONDecodable, A == A.DecodedType>(value: JSON) -> Parser<[String: A]> {
+  switch value {
+  case let .Object(o): return sequence(map(o, A.fromJSON))
```
gfontenot: the partial application of `A.fromJSON` might cause segfaults on 32 bit systems right now. Do we care? We aren't cutting a new release till 1.2 is out, at which point that bug should be fixed anyway.

Cartfile.resolved
```diff
@@ -1 +1,2 @@
+github "robrix/Box" "1.1.0"
```
gfontenot: We should probably be more generic here and say `">= 1.1.0"`

gfontenot: Actually, this shouldn't have quotes at all. And it should be `~> 1.1.0`

Cartfile.resolved
```diff
@@ -1 +1,2 @@
...
 github "thoughtbot/runes" "v1.2.0"
```
gfontenot: This should also have no quotes, and should be `~> 1.2.0` (I know this wasn't your change, but since you're in here)

Cartfile.resolved
gfontenot: I'm making comment sin the `Cartfile.resolved` because I'm a dummy. Ignore me.

Argo/Globals/Parser.swift
gfontenot: Personally, I'd rather `optional` since I don't get that this is doing any kind of parsing.

Argo/Globals/Parser.swift
```diff
@@ -0,0 +1,22 @@
...
+
+  public var value: T? {
+    switch self {
+    case let .Success(box): return box.value
+    default: return .None
+    }
+  }
+}
+
+func optional<A>(p: Parser<A>) -> Parser<A?> {
```
gfontenot: Sorry, I do think this should be a static function on `Parser` to avoid polluting the global scope.

jferris: Is this private? If so, isn't the "global" scope just within Argo?

Argo/Globals/StandardTypes.swift
```diff
@@ -2,53 +2,64 @@ import Foundation
...
-  case let .Array(a): return sequence({ A.decode($0) } <^> a)
-  default: return .None
+  case let .Array(a): return sequence({ A.fromJSON($0) } <^> a)
+  default: return typeMismatch("Array", value)
   }
 }
+
+public func decodeObject<A where A: JSONDecodable, A == A.DecodedType>(value: JSON) -> Parser<[String: A]> {
+  switch value {
+  case let .Object(o): return sequence(map(o) { A.fromJSON($0) })
```
gfontenot: Is it weird that this is now using a different syntax than `decodeArray`, which is doing the exact same thing? Maybe we should change `decodeArray` to match this instead of defining `<^>` for dictionaries?

ArgoTests/Tests/ExampleTests.swift
```diff
@@ -13,7 +12,7 @@ class ExampleTests: XCTestCase {
 
   func testJSONWithRootObject() {
     let json = JSON.parse <^> JSONFileReader.JSON(fromFile: "root_object")
-    let user: User? = json >>- { $0 <| "user" >>- User.decode }
+    let user: User? = json >>- { ($0 <| "user").value }
```
gfontenot: it's a shame that `.value` gets introduced here. There's no way to avoid that?

Argo/Globals/Parser.swift
gfontenot: It's internal, so kinda? I do think attaching it to `Parser` still makes sense. The syntax where we're using it won't change much (`optional(foo)` -> `.optional(foo)`)

Given that it's internal, I can let go of this, though.

ArgoTests/Tests/ExampleTests.swift
tonyd256: I would like to use the decode function but as it is now, it can't take out root keys ... we might possibly want to make decode functions that also take JSON in instead of AnyObject

gfontenot: That probably doesn't need to happen for this PR, but it would be nice to look into.

gfontenot:
I'm really happy with how this turned out. Fantastic work. I think we should do this. Any comments/concerns from @thoughtbot/ios (or anyone else watching this repo) here?

tonyd256:
Also, a question on naming. I've used `fromJSON` in place of the old `decode` function but Aeson uses `FromJSON` as the protocol name and `parseJSON` as the function name. If our function is returning a `Parser` object then does `parseJSON` make more sense for us too?

jferris:
I think it's cool to use Aeson as a model, but I think you should feel free to use what you think is intuitive and will generally make sense to the iOS community Argo is intended for. I'm guessing you'll have more users who have previously used Objective-C than have previously used Haskell. I'd pick names that will work for your audience.

tonyd256:
I agree with you Joe, but I don't think any of these names would be confusing to iOS devs. I'm open to anything though, not just `fromJSON` or `parseJSON`.

gfontenot:
Yeah, either would really work. I'd be interested in feedback from @thoughtbot/ios on this.

nvh:
I really like this solution, but I think the name `Parser` is a bit confusing. It's just a value type and not doing much parsing, like the name suggests. Wouldn't it be better to call it `Result` or `ParseResult`?

Also, I think it would be nice to have a constructor for `Parser` that takes a `T`, or a global `pure()` function, so we can forget about the boxing if we create a `Parser`.

hyperspacemark:
> It's just a value type and not doing much parsing, like the name suggests. Wouldn't it be better to call it `Result` or `ParseResult`?

+1 for `ParseResult`.

tonyd256:
I like ParseResult as well ... made those changes.   Does it make sense now to use `parseJSON` instead of `fromJSON`? I think I like `func parseJSON(JSON) -> ParseResult<User>` better than `func fromJSON(JSON) -> ParseResult<User>`

hyperspacemark:
> I like ParseResult as well ... made those changes.

Don't forget to rename `Parser.swift` to `ParseResult.swift`.

> Does it make sense now to use `parseJSON` instead of `fromJSON`?

We're talking about the function that exists on `JSON`, right? If so, I'd argue that that function should just be called `parse()`. Following this line of reasoning, maybe `JSONDecodable` deserves a new name as well? `JSONParsable`? Just `Parsable`?

gfontenot:
> We're talking about the function that exists on JSON, right? If so, I'd argue that that function should just be called parse(). Following this line of reasoning, maybe JSONDecodable deserves a new name as well? JSONParsable? Just Parsable?
> 
He's talking about the function on `JSONDecodable` instances.

nvh:
> I think I like func `parseJSON(JSON) -> ParseResult<User>` better than func `fromJSON(JSON) -> ParseResult<User>`

I also think the first one is better, but agree with @hyperspacemark that the protocol name `JSONDecodable` is weird in that case. On the other hand, I like `JSONDecodable` much better that `JSONParsable`

Just another suggestion: the name parsing is used convert from strings or AnyObject to structured data, wether it be to JSON or some model. the name decoding is used when converting JSON to some model. The result type can than be called `DecodeResult` and the global functions would then be called `parse`:
```swift
public func parse<T: JSONDecodable where T == T.DecodedType>(object: AnyObject) -> DecodeResult <T>
```
So the protocol would be
```swift
public protocol JSONDecodable {
  typealias DecodedType = Self
  static func decode(json: JSON) -> DecodeResult<DecodedType>
}
```

tonyd256:
Tried to name things more like `decode` since the protocol is `JSONDecodable`. Take a look and leave feedback. I can always drop this commit if it's not as nice. @thoughtbot/ios @nvh

nvh:
This is indeed exactly as I had it in mind.

gfontenot:
I'm cool with this. Do we need to name `parse` and `decode` differently? Is there any real benefit there?

tonyd256:
@thoughtbot/ios Any more feedback on this? I'm going to merge it in by end of day if not.

Argo.xcodeproj/project.pbxproj
```diff
@@ -26,6 +30,10 @@
 		EABDF6921A9CD4EA00B6CC83 /* types.plist in Resources */ = {isa = PBXBuildFile; fileRef = EABDF68E1A9CD4EA00B6CC83 /* types.plist */; };
 		EABDF6941A9CD4FC00B6CC83 /* PListDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EABDF6931A9CD4FC00B6CC83 /* PListDecodingTests.swift */; };
 		EABDF6951A9CD4FC00B6CC83 /* PListDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EABDF6931A9CD4FC00B6CC83 /* PListDecodingTests.swift */; };
+		EAC59A0F1A9F74C500BEB5CC /* DecodeResultOperators.swift in Sources */ = {isa = PBXBuildFile; fileRef = EAC59A0E1A9F74C500BEB5CC /* DecodeResultOperators.swift */; };
```
gfontenot: We should just call this `DecodeOperators.swift`

Argo/Globals/StandardTypes.swift
```diff
@@ -2,53 +2,64 @@ import Foundation
...
     }
   }
 }
 
-public func decodeArray<A where A: JSONDecodable, A == A.DecodedType>(value: JSON) -> [A]? {
+public func decodeArray<A where A: JSONDecodable, A == A.DecodedType>(value: JSON) -> Decoded<[A]> {
   switch value {
-  case let .Array(a): return sequence(A.decode <^> a)
-  default: return .None
+  case let .Array(a): return sequence(a.map { A.decode($0) })
```
gfontenot: Both of these can be `sequence(A.decode <^> a)` now. Might as well make that change while we're futzing around in here.

Argo/Operators/JSONOperators.swift
```diff
@@ -3,46 +3,44 @@ import Runes
...
-  return pure(json <| key)
+public func <|?<A where A: JSONDecodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<A?> {
+  return .optional(json <| key)
 }
 
 // Pull embedded value from JSON
-public func <|<A where A: JSONDecodable, A == A.DecodedType>(json: JSON, keys: [String]) -> A? {
-  return flatReduce(keys, json, <|) >>- A.decode
+public func <|<A where A: JSONDecodable, A == A.DecodedType>(json: JSON, keys: [String]) -> Decoded<A> {
+  return flatReduce(keys, json, <|) >>- { A.decode($0) }
```
gfontenot: This can use partial application now

Cartfile
```diff
@@ -1 +1,2 @@
 github "thoughtbot/runes" ~> 1.2
+github "robrix/Box" ~> 1.1.0
```
gfontenot: want to bump this to `1.2.0`?

gfontenot:
This is really great now. Big fan.

PR #77, opened 2015-03-06
Add carthage compatible badge

tonyd256:
Want to add this in. The placement is up for discussion. Any opinions?


gfontenot:
The convention I've seen is to put it at the very top of the README.

tonyd256:
I've seen that as well but also seen things mixed in so ya

tonyd256:
@gfontenot is this better placement?

gfontenot:
I like that better, yeah. Want to open one of these for Runes as well?

PR #78, opened 2015-03-11
[README] Update CocoaPods instructions

kylef:
0.36 shipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,12 +39,10 @@ dependency of Argo, so you don't need to specify it in your Cartfile.
 
 [CocoaPods]: http://cocoapods.org
 
-__DISCLAIMER: CocoaPods doesn't officially support Swift projects yet. Use the
-pre-release version at your own discretion.__
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,12 +39,10 @@ dependency of Argo, so you don't need to specify it in your Cartfile.
 
 [CocoaPods]: http://cocoapods.org
 
-__DISCLAIMER: CocoaPods doesn't officially support Swift projects yet. Use the
```
hyperspacemark: I think we should keep the disclaimer and change it to make it very, very clear that you need version 0.36.0 of Cocoapods otherwise you're going to have a bad time. Then we can remove the line below about installing with that particular version.

kylef: I wouldn't say it's that important. In a few days time the minimum version of CocoaPods permitted with the master specs repository will be bumped to 0.36. Due to numerous backwards incompatible changes in pod specs.

https://github.com/CocoaPods/Specs/blob/master/CocoaPods-version.yml#L2

gfontenot:
Thanks! Closing this in favor of the wording in #85

PR #83, opened 2015-03-13
Remove closures in favor of partial application

gfontenot:
There was a bug in Swift before version 1.2 beta 3 that meant we had to
wrap these partial application sections in a closure in order to avoid
segfaults. As of 1.2 beta 3, this bug in Swift has been fixed, so we can
remove the explicit closures and just apply the functions the way god
intended.


iancanderson:
:heart_eyes_cat:

tonyd256:
such a small change yet so amazing :+1:

PR #88, opened 2015-03-17
Improve `JSON.parse` performance

tonyd256:
Using the update/append functionality of Dictionaries instead of
implementing a custom merge operator `+`, we see significant speed
increases for large data sets. Given a ~85K json file, it took ~265ms to
parse without this optimization and ~75ms with this optimization.

### DIFF
--- a/Argo/Globals/JSON.swift
+++ b/Argo/Globals/JSON.swift
@@ -15,9 +15,10 @@ public extension JSON {
     case let v as [AnyObject]: return .Array(v.map(parse))
 
     case let v as [Swift.String: AnyObject]:
-      return .Object(reduce(v.keys, [:]) { accum, key in
+      return .Object(reduce(v.keys, [:]) { (var accum, key) in
         let parsedValue = (self.parse <^> v[key]) ?? .Null
-        return accum + [key: parsedValue]
+        accum[key] = parsedValue
+        return accum
       })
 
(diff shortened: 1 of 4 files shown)


gfontenot:
:+1:

gfontenot:
Actually, we should remove our implementation of `+` for Dictionaries if we aren't using it anymore. Then :+1:

Argo/Globals/JSON.swift
```diff
@@ -15,9 +15,10 @@ public extension JSON {
     case let v as [AnyObject]: return .Array(v.map(parse))
 
     case let v as [Swift.String: AnyObject]:
-      return .Object(reduce(v.keys, [:]) { accum, key in
-        let parsedValue = (self.parse <^> v[key]) ?? .Null
-        return accum + [key: parsedValue]
+      return .Object(reduce(v, [:]) { (var accum, elem) in
+        let parsedValue = (self.parse <^> elem.1) ?? .Null
+        accum[elem.0] = parsedValue
```
jferris: Does this mutate `v`?

tonyd256: only `accum` is being mutated

gfontenot: But `accum` is a value object, so it's passed by copy. It's only mutating the instance in the local scope.

PR #96, opened 2015-03-27
Add .fromOptional to Decoded

gfontenot:
`.fromOptional` converts `T?` into `Decoded<T>`, which makes it easier
to `flatMap` the `AnyObject?` returned from `NSJSONSerialization` into
`decode` and get back a `Decoded<T>`.

### DIFF
--- a/Argo/Types/Decoded.swift
+++ b/Argo/Types/Decoded.swift
@@ -12,12 +12,21 @@ public enum Decoded<T> {
     default: return .None
     }
   }
+}
 
-  public static func optional<A>(p: Decoded<A>) -> Decoded<A?> {
-    switch p {
+public extension Decoded {
+  static func optional<T>(x: Decoded<T>) -> Decoded<T?> {
+    switch x {
     case let .Success(box): return .Success(Box(.Some(box.value)))
     case let .MissingKey(string): return .Success(Box(.None))
     case let .TypeMismatch(string): return .TypeMismatch(string)
+    }
+  }
+
+  static func fromOptional<T>(x: T?) -> Decoded<T> {
+    switch x {
+    case let .Some(value): return .Success(Box(value))
+    case .None: return .TypeMismatch("Expected .Some(\(T.self)), got .None")
     }
   }
 }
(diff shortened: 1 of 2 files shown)


Argo/Types/Decoded.swift
```diff
@@ -12,14 +12,23 @@ public enum Decoded<T> {
     default: return .None
     }
   }
+}
 
-  public static func optional<A>(p: Decoded<A>) -> Decoded<A?> {
-    switch p {
+public extension Decoded {
+  static func optional<T>(x: Decoded<T>) -> Decoded<T?> {
```
tonyd256: This function isn't the inverse of `fromOptional` does that seem weird? Should we have a different name for this?

hyperspacemark: I think this name makes perfect sense. I'd expect the inverse of `fromOptional()` to be `toOptional()`.

tonyd256: I agree the `toOptional` sounds like it's inverse but do you think that calling `optional` on `Decoded<T>` would return `Decoded<T?>` ... I don't think I would infer that from the name. I don't have a better idea though so I'm ok with it if everyone else is.

hyperspacemark: I could also go either way on this one. Just out of curiosity, what would you expect a function named `optional` to return?

gfontenot: I think that the type signature makes that clear, and the name itself makes a lot of sense in use.

tonyd256: I would expect `optional` to return an optional honestly. The convention I've seen with iOS is to have the type you want there ... kinda like `UIColor.whiteColor().CGColor` so in this case I might expect `.Optional` to return an optional form. This function is `.optional` so it's lower case but I think it's close enough to be slightly confusing.

hyperspacemark: Oh I see your point now. When I first read the return type I read it as `Decoded<T>?` so I was confused by your confusion about the name. Now I see that it's not an optional decoded T, it's a decoded optional T.

gfontenot: I think it's really important to see this in context https://github.com/thoughtbot/Argo/blob/master/Argo/Operators/Operators.swift#L16-L18

tonyd256: I don't have a better name to recommend, I'm ok with this and was just seeing if other people had ideas for it.

tonyd256:
:+1:

PR #100, opened 2015-03-31
Add 'import Runes' to README tl;dr

exclsr:
Add import statements to the tl;dr, as the map and apply operators are a part of Runes and require more than "import Argo" to compile. It took me a while to figure this out.


gfontenot:
Thanks!

PR #107, opened 2015-04-15
Rename JSONDecodable to Decodable

gfontenot:
`Decodable` objects are able to be `decode`d, returning a
`Decoded<DecodedType>`. This is a nice synergy, and leaves us open to
other non-JSON operations in the future.

### DIFF
--- a/Argo/Functions/decode.swift
+++ b/Argo/Functions/decode.swift
@@ -1,15 +1,15 @@
-public func decode<T: JSONDecodable where T == T.DecodedType>(object: AnyObject) -> T? {
+public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) -> T? {
   return decode(object).value
 }
 
-public func decode<T: JSONDecodable where T == T.DecodedType>(object: AnyObject) -> [T]? {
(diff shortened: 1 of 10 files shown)


tonyd256:
I like the change in general because I think it would be nice to get away from JSON references in general. This is getting us one step closer. Using `Value` instead of `JSON` would be the final nail in the coffin.

I'm a :+1: for this!

gfontenot:
I think right now, `JSON` is at least conceptually tied to parsing JSON. If an example comes up where we (or someone else) has, for example, a YAML parser that works with the `JSON` type without modification, I think it would make sense to change it at that point. And we could always `typealias JSON = Value` if it comes to that to maintain backwards compatibility.

gfontenot:
ping @thoughtbot/ios and @hyperspacemark for feedback please

dazmuda:
:+1: I really like this change. You `decode` `Decodable` objects. Even when parsing JSON, the nomenclature `decode`ing `JSONDecodable` objects is a bit awkward.

hyperspacemark:
I dig it. :+1:

jnutting:
I am all in favor of this :shipit:

Issue #106, opened 2015-04-12
Expression too long issues in Xcode 6.3 / Swift 1.2


czeluff:
Apologies for the enormous code block - figure I should just show this in its entirety. I've seen previous issues related to this need to break up the object into multiple pieces to make the 'decode' function work properly. However, I also appear to be having an issue with the curried 'create' function - it seems to be giving up around "item3" due to too many parameters.

Some people have shown potential fixes by breaking up 'decode' [like so](https://github.com/thoughtbot/Argo/issues/5#issuecomment-78961913) but this is not compiling for me, either.

If I need to break up the object into multiple pieces, I'll have to. Though this can be pretty annoying, requiring more curried 'create' functions.

```swift
public struct RawStats {
    let assists: Int
    let barracksKilled: Int
    let championsKilled: Int
    let combatPlayerScore: Int
    let consumablesPurchased: Int
    let damageDealtPlayer: Int
    let doubleKills: Int
    let firstBlood: Int
    let gold: Int
    let goldEarned: Int
    let goldSpent: Int
    let item0: Int
    let item1: Int
    let item2: Int
    let item3: Int
    let item4: Int
    let item5: Int
    let item6: Int
    let itemsPurchased: Int
    let killingSprees: Int
    let largestCriticalStrike: Int
    let largestKillingSpree: Int
    let largestMultiKill: Int
    let legendaryItemsCreated: Int
    let level: Int
    let magicDamageDealtPlayer: Int
    let magicDamageDealtToChampions: Int
    let magicDamageTaken: Int
    let minionsDenied: Int
    let minionsKilled: Int
    let neutralMinionsKilled: Int
    let neutralMinionsKilledEnemyJungle: Int
    let neutralMinionsKilledYourJungle: Int
    let nexusKilled: Bool
    let nodeCapture: Int
    let nodeCaptureAssist: Int
    let nodeNeutralize: Int
    let nodeNeutralizeAssist: Int
    let numDeaths: Int
    let numItemsBought: Int
    let objectivePlayerScore: Int
    let pentaKills: Int
    let physicalDamageDealtPlayer: Int
    let physicalDamageDealtToChampions: Int
    let physicalDamageTaken: Int
    let playerPosition: Int
    let playerRole: Int
    let quadraKills: Int
    let sightWardsBought: Int
    let spell1Cast: Int
    let spell2Cast: Int
    let spell3Cast: Int
    let spell4Cast: Int
    let summonSpell1Cast: Int
[...]

PR #110, opened 2015-04-15
Adding specific Int64 type support 

MichaelMcGuire:
On 32-bit platforms (namely iPhone 4S, iPhone 5 for iOS8+) the Swift Int type only goes up to 2^31.  Even though there is technically no limits placed upon JSON numeric values, they are often designed around JavaScript clients which have a maximum integer value of 2^53.  I've added explicit support for the Int64 type to support this limit.  

In order to get around Xcode 6.3/Swift 1.2 (see #106) limitations, I split out the numeric tests into a sub-model.  Using the solution presented in #5 didn't seem to work.


MichaelMcGuire:
Probably important to mention that if a numeric value of greater than 2^31 is present in the data and you are on a 32-bit device using a plain Int type, you get a silent overflow on decoding.

gfontenot:
Killer. I'm going to reword your commit message to include some of the info you've provided here and then merge it in.

gfontenot:
Done in ddee5da

MichaelMcGuire:
Awesome, thanks!

PR #111, opened 2015-04-16
Update framework search paths.

paulyoung:
This pull request removes entries in `Framework Search Paths` that reference the `Carthage/Build` directories.

I'm using Carthage with the `--no-build` and `--use-submodules` flags and I'm getting `Directory not found` warnings when building because of those.

They may be intentional but I suspect they're as a result of http://www.openradar.me/19124221.

In any case, I figured a pull request was the best way to communicate the problem I was having.

### DIFF
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -834,11 +834,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
-				FRAMEWORK_SEARCH_PATHS = (
-					"$(inherited)",
-					"$(PROJECT_DIR)/Carthage/Build/iOS",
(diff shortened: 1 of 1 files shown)


gfontenot:
That radar isn't exactly the same issue. We're intentionally adding `Carthage/Build/*` to the framework search paths, because that's needed for Xcode to be able to find our dependencies. Unfortunately, removing those settings breaks the build on my machine.

Can you explain more about how you're trying to use this project so that I have a clearer idea on why this is causing issues for you?

paulyoung:
I have a workspace which includes a project with framework targets for Mac and iOS.

By using the `--no-build` and `--use-submodules` flags my workflow is the same as using submodules without Carthage; adding projects to my workspace and then linking to them.

I've set up a quick project at https://github.com/paulyoung/ArgoCarthageIssue which also displays these warnings.

paulyoung:
@tonyd256 any update?

tonyd256:
Hey sorry about that. I deleted that branch and it closed this. I'm going to open this back up as an issue.

PR #115, opened 2015-04-27
Update podspec for 1.0.0

tonyd256:


Argo.podspec
```diff
@@ -17,5 +17,6 @@ Pod::Spec.new do |spec|
   spec.osx.deployment_target = '10.9'
 
   spec.dependency 'Runes'
+  spec.dependency 'Box'
```
gfontenot: We should pin these to the Swift 1.2 versions:

`spec.dependency 'Runes', '~> 1.2'` (I think)

Box should be ~> 1.2.0 as well.

gfontenot:
:+1:

PR #117, opened 2015-04-28
Add a function to convert JSON to .None if = .Null

tonyd256:
With the addition of the `Decoded` type we decided to return a Type
Mismatch whenever the type asked for did not align with the `JSON`. In
the case of Null, we want to assume that it means that the key is just
missing and not of the wrong type. This adds code to convert the `JSON`
into an optional and specifically return `.None` when the `JSON` ==
`.Null`.

### DIFF
--- a/ArgoTests/JSON/types.json
+++ b/ArgoTests/JSON/types.json
@@ -11,6 +11,7 @@
     "hello",
     "world"
   ],
+  "string_array_opt": null,
   "embedded": {
     "string_array": [
       "hello",
--- a/Argo/Operators/Operators.swift
+++ b/Argo/Operators/Operators.swift
@@ -9,7 +9,7 @@ infix operator <||? { associativity left precedence 150 }
 
 // Pull value from JSON
 public func <|<A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<A> {
-  return decodeObject(json) >>- { (A.decode <^> $0[key]) ?? .MissingKey(key) }
+  return decodeObject(json) >>- { (A.decode <^> $0[key] >>- JSON.optional) ?? .MissingKey(key) }
 }
 
 // Pull optional value from JSON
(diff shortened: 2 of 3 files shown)


tonyd256:
@thoughtbot/ios thoughts here? I'd like to get this in asap

gfontenot:
What if we used `.Null` and something akin to Haskell's `guard`?

```swift
decodeObject(json) >>- { pure($0[key] ?? .Null) } >>- guardNull(key) >>- A.decode

func guardNull(key: String)(j: JSON) -> Decoded<JSON> {
    switch j {
    case .Null: return .MissingKey(key)
    default: return .Success(j)
    }
}
```

ArgoTests/JSON/types.json
```diff
@@ -11,6 +11,7 @@
     "hello",
     "world"
   ],
+  "string_array_opt": null,
```
gfontenot: We should actually try decoding this into something in the tests too. As it is, we never touch this value.

tonyd256: It's being decoded in the `TestModel` decode function. then we're testing that it's nil in the TypeTests file. Should we add a test that gets the `Decoded<TestModel>` and checks if that is `.MissingKey()`?

What we have now should be testing that issue.

gfontenot: Ah, no, I missed the fact that we were already trying to pull this key. Carry on.

gfontenot:
I really like this. Maybe @jferris can comment on if we're over complicating it?

jferris:
The semantics of "missing key" vs "explicit null" are interesting. Have you played with how this interacts with optional fields? For example, if a a user has an optional email, and the email key is explicitly set to `null` rather than being omitted, does that work as expected?

gfontenot:
@jferris That's what the tests are doing now. With this change, all missing keys are actually just turned into `.Null` internally, which becomes `.MissingKey`. So this _removes_ any semantic difference between `null` and just omitting the key that may have existed previously.

PR #120, opened 2015-05-01
Bring back CI with Circle-CI

tonyd256:


gfontenot:
...how does this work without Carthage?

gfontenot:
Oh, the carthage stuff is set up on Circle itself. That weirds me out.

tonyd256:
@gfontenot I agree that I would like to be using bin/setup here but Circle-ci is having the same damn permissions issues that mark was having with carthage ... so I forced it to compile the Mac side only so it will work

gfontenot:
:+1:

Issue #121, opened 2015-05-01
framework search paths issue when using submodules


tonyd256:
When using the Carthage flag `--use-submodules` and `--no-build` in a project that depends on Argo, Argo's framework search paths do not exist because they point to the Carthage build directory. However, removing these breaks the build when not using those flags.

This issue can be a discussion on how to fix this.

@paulyoung Please add any info to this issue that I might have missed above.

paulyoung:
We can always refer to #111 and this project: https://github.com/paulyoung/ArgoCarthageIssue

gfontenot:
This should be fixed with #123.

PR #122, opened 2015-05-03
implement native submodules instead of requiring external dep manager

sebbean:
seems strange to require someone to use a package manager when submodules are baked right into git. 
They're pretty darn simple once you start using them.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "Vendor/Runes"]
+	path = Vendor/Runes
+	url = https://github.com/thoughtbot/Runes.git
+[submodule "Vendor/Box"]
+	path = Vendor/Box
+	url = https://github.com/robrix/Box.git
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -104,6 +104,62 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
+		964283E61AF6AE9500C586FC /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 964283DF1AF6AE9500C586FC /* Box.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = D470AC3719E86128003DA6C6;
+			remoteInfo = "Box-Mac";
+		};
+		964283E81AF6AE9500C586FC /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 964283DF1AF6AE9500C586FC /* Box.xcodeproj */;
+			proxyType = 2;
+			remoteGlobalIDString = D470AC4219E86128003DA6C6;
+			remoteInfo = "Box-MacTests";
+		};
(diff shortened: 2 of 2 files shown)


kaishin:
> They're pretty darn simple once you start using them.

I couldn't disagree more. Never had a fun time using submodules. Package managers are the proper way to deal with software dependencies because they abstract the git implementation details and encourage authors to use semantic versioning.

gfontenot:
Agree with @kaishin. We don't require Carthage to _use_ Argo, which is the important thing. Development dependencies are development dependencies. There's nothing strange about using them. I very much dislike git submodules, and have no interest in (re-)introducing them here.

sebbean:
You do require *something* to pull in Runes and Box. Otherwise Argo will not build.

sebbean:
yea, i guess what's great about submodules is you can very easily create pull requests for the projects that are deps and fix code that may need fixing. 
creates a very clean flow to push upstream to external open source projects.

it's really just git inside git.

i was all about dep managers up until cocoapods fell apart. carthage wasn't really doing anything for anyone.. just adding confusion. thought it was about time to figure out a real solution.

not that this is the venue but i'd love to hear some of your issues with them. i find it as a great way to deal with interdependent projects while keeping them in separate repo.

sebbean:
curious what you mean by not *requiring* carthage since there's no other way to pull in the dependencies that Argo requires. @gfontenot

gfontenot:
@sebbean We support Carthage and CocoaPods as client-side dependency management solutions, as well as provide prebuilt binaries for the dependencies we control (Argo and Runes). Box is also offered as a prebuilt binary. Our support of git submodules as an installation method is incidental (and [kind of broken][#121]). I'm interested in finding a way to support git submodules as a dependency solution, but not incredibly interested in bloating our repo and complicating our development process to accommodate it.

[#121]: https://github.com/thoughtbot/Argo/issues/121

gfontenot:
Maybe I'm being too dogmatic about this. I'm going to re-open for now. I'll continue thinking about this and we can continue the discussion.

gfontenot:
I'm closing this again in favor of the more comprehensive change in #123. We will use git submodules moving forward, assuming that PR gets merged.

sebbean:
thanks for the reconsideration @gfontenot 

funnily enough i see git submodules as significantly less complicated than attempting to learn and understand what another dep manager is attempting to do and how it's doing it.

i spent a fair amount of time with carthage after realizing cocoapods wasn't up to the task (few months back before they figured it out). I realized that I actually had no idea what any of these things were actually doing and I was relying too much on them to put my project / workspaces together. 

Since moving to submodules, I've allowed myself to work through the process of adding any type of dependencies to my project without the reliance on another system. I now am much more comfortable debugging framework / library related issues that were completely beyond me in the days of blind cocopods usage.

Not to mention the amount of PR's I've created with minor fixes to open source projects (dependencies) since the workflow is painless. Takes me 2 minutes to fix a lil warning or bug and push up to be reviewed. This PR included!

I think that, although it may not be for everyone, becoming more comfortable with this flow would not only help end users understand how projects *actually* interact with each other, it would also significantly increase end user contributions to projects that they are dependent on.

PR #123, opened 2015-05-04
Use Git Submodules for development

### LINKED ISSUE #121: framework search paths issue when using submodules

gfontenot:
This solves a couple of issues:

1. Users that pass `--no-build --use-submodules` can now use Argo
   without running into errors from missing framework search paths
   (fixes #121)
2. Users that want to use Submodules as a dependency management solution
   can now do so without having to jump through any additional hoops.

We're still going to use Carthage as a development dependency, but it's
going to act as a semantic wrapper around submodules, instead of relying
on the binaries.

I'm not super happy about this change, but it seems like it's the most
pragmatic way to move forward. I really wish git submodules would just
go the fuck away.

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,2 +1,2 @@
-github "robrix/Box" "1.2.0"
+github "robrix/Box" "1.2.1"
 github "thoughtbot/runes" "v1.2.2"
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,7 @@ build/
 Pods
 
 # Carthage
-Carthage/
+Carthage/Build
 
 # AppCode specific files
 .idea/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "Carthage/Checkouts/Box"]
+	path = Carthage/Checkouts/Box
+	url = https://github.com/robrix/Box.git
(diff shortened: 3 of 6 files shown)


gfontenot:
Interestingly enough, the move to a workspace means we might be able to add a playground directly to this project, which would be nice for educational purposes.

gfontenot:
ping @thoughtbot/ios on this

Cartfile.resolved
```diff
@@ -1,2 +1,2 @@
-github "robrix/Box" "1.2.0"
+github "robrix/Box" "1.2.1"
 github "thoughtbot/runes" "v1.2.2"
```
dazmuda: Is there a reason we aren't bumping this to Runes 2.0?

gfontenot: We'll do that along with #125

dazmuda: Rockin. Sorry, didn't see that one.

tonyd256:
:+1:

PR #125, opened 2015-05-05
Remove version specification for Runes

gfontenot:
We don't actually need to be so prescriptive about what version of Runes
we're using anymore. Any version after we turned on extension
compatibility will be fine for our needs.


tonyd256:
:+1:

PR #128, opened 2015-05-08
Add printable to decoded type

tonyd256:
This adds printable to `Decoded` for easily passing the decoded model into a
`println` statement for debugging.

### DIFF
--- a/Argo/Types/Decoded.swift
+++ b/Argo/Types/Decoded.swift
@@ -31,6 +31,16 @@ public extension Decoded {
   }
 }
 
+extension Decoded: Printable {
+  public var description: String {
+    switch self {
+    case let .Success(x): return "Success(\(x))"
+    case let .TypeMismatch(s): return "TypeMismatch(\(s))"
+    case let .MissingKey(s): return "MissingKey(\(s))"
+    }
+  }
+}
+
 public extension Decoded {
   func map<U>(f: T -> U) -> Decoded<U> {
     switch self {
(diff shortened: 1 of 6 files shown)


Argo/Types/Decoded.swift
```diff
@@ -31,6 +31,16 @@ public extension Decoded {
   }
 }
 
+extension Decoded: Printable {
+  public var description: String {
+    switch self {
+    case let .Success(x): return x.description
+    case let .TypeMismatch(s): return s
+    case let .MissingKey(s): return s
```
gfontenot: Do we want these values wrapped in their enum case? That would mimic the behavior of `Optional`, `Result`, etc.

jnutting: +1

tonyd256: :+1:

tonyd256:
@thoughtbot/ios Any more thoughts?

Argo/Types/Decoded.swift
```diff
@@ -31,6 +31,16 @@ public extension Decoded {
   }
 }
 
+extension Decoded: Printable {
+  public var description: String {
+    switch self {
+    case let .Success(x): return "Success(\(x.description))"
```
gfontenot: Do we need to do `x.description` here? Can we not just use `x`?

tonyd256: oh right .. that should work

tonyd256:
@gfontenot good to go?

gfontenot:
:+1:

PR #130, opened 2015-05-08
Add playground

gfontenot:
Now that we're using git submodules (grumble grumble) and a workspace,
we can add a playground to the workspace to act as a scratch
pad/educational document for users.

### DIFF
--- a/Argo.playground/playground.xcworkspace/contents.xcworkspacedata
+++ b/Argo.playground/playground.xcworkspace/contents.xcworkspacedata
@@ -2,6 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "group:../../Code/thoughtbot/Argo/Argo.playground">
+      location = "group:Argo.playground">
    </FileRef>
 </Workspace>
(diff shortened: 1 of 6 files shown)


Argo.playground/playground.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "group:../../Code/thoughtbot/Argo/Argo.playground">
```
tonyd256: not sure this will work

tonyd256:
1 comment that should be dealt with but otherwise :+1: 

I think this is a really great addition!

Argo.playground/playground.xcworkspace/contents.xcworkspacedata
gfontenot: Oh, huh. I don't know where this path is coming from...

PR #133, opened 2015-05-21
Replace decodeObject with json object direct lookup

tonyd256:
Using `decodeObject` on the JSON in the `<|` operator is causing a heavy
performance hit. This operator is the base operator that all the other ones
rely on. That means that this one is called every time any other one it
called. The `decodeObject` function is looping through every key-value pair
in the object and calling `decode` on the value. Then after this function
returns we are using the dictionary subscript to pull only one object from
that dictionary, meaning we just decoded a bunch of values that we didn't
need to decode. This unused decoding is happening so often that we're
seeing performance hits of 100 times or more.

Pulling out the interal JSON from the given JSON first allows us to decode
only what we're looking for each time.

### DIFF
--- a/Argo/Operators/Operators.swift
+++ b/Argo/Operators/Operators.swift
@@ -9,7 +9,7 @@ infix operator <||? { associativity left precedence 150 }
 
 // Pull value from JSON
 public func <|<A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<A> {
-  return decodeObject(json) >>- { pure($0[key] ?? .Null) } >>- guardNull(key) >>- A.decode
+  return decodedJSONForKey(json, key) >>- A.decode
 }
 
 // Pull optional value from JSON
@@ -19,7 +19,7 @@ public func <|?<A where A: Decodable, A == A.DecodedType>(json: JSON, key: Strin
 
 // Pull embedded value from JSON
 public func <|<A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [String]) -> Decoded<A> {
-  return flatReduce(keys, json, <|) >>- A.decode
+  return flatReduce(keys, json, decodedJSONForKey) >>- A.decode
 }
 
 // Pull embedded optional value from JSON
@@ -47,11 +47,4 @@ public func <||<A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [Str
 // Pull embedded optional array from JSON
 public func <||?<A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [String]) -> Decoded<[A]?> {
   return .optional(json <|| keys)
-}
-
-private func guardNull(key: String)(j: JSON) -> Decoded<JSON> {
-  switch j {
-  case .Null: return .MissingKey(key)
-  default: return pure(j)
-  }
 }
--- a/Argo/Types/StandardTypes.swift
+++ b/Argo/Types/StandardTypes.swift
@@ -69,6 +69,20 @@ public func decodeObject<A where A: Decodable, A == A.DecodedType>(value: JSON) 
   }
 }
 
+func decodedJSONForKey(json: JSON, key: String) -> Decoded<JSON> {
+  switch json {
+  case let .Object(o): return guardNull(key, o[key] ?? .Null)
+  default: return typeMismatch("Object", json)
+  }
+}
+
 private func typeMismatch<T>(expectedType: String, object: JSON) -> Decoded<T> {
   return .TypeMismatch("\(object) is not a \(expectedType)")
+}
+
+private func guardNull(key: String, j: JSON) -> Decoded<JSON> {
+  switch j {
+  case .Null: return .MissingKey(key)
+  default: return pure(j)
+  }
 }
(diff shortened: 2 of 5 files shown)


hyperspacemark:
@tonyd256 Wow that's awesome. Would you mind publishing the `measureBlock` output?

tonyd256:
output:
```
Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' started.
Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' measured [Time, seconds] average: 0.021, relative standard deviation: 7.125%, values: [0.023963, 0.021785, 0.021894, 0.019390, 0.019479, 0.019491, 0.019371, 0.019340, 0.020768, 0.020373], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' passed (0.720 seconds).
Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' started.
Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' measured [Time, seconds] average: 0.044, relative standard deviation: 4.891%, values: [0.045203, 0.043861, 0.046231, 0.044899, 0.039911, 0.043974, 0.044074, 0.043960, 0.047325, 0.040639], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' passed (0.699 seconds).
```

hyperspacemark:
Cool, what was it before?

jnutting:
I just ran this test on master, where I got 1.57s for testDecodePerformance, and 0.06s for testParsePerformance. Running on this branch, I got 0.04 and 0.05 respectively, so it really looks like a huge win for decoding here! Roughly a 40x speedup.

tonyd256:
Any comments on the code? It definitely improves performance substantially as seen here and in #132 so if there are no code comments I'd like to get this in today. @thoughtbot/ios @hyperspacemark

Argo/Types/JSON.swift
```diff
@@ -28,6 +28,15 @@ public extension JSON {
   }
 }
 
+extension JSON {
+  func jsonForKey(key: Swift.String) -> JSON {
```
jferris: I feel like getting "object" into the name for this function would be useful. Maybe `parseObject` or something like that? It's like `decodeObject` except it doesn't follow through and map out the keys.

Argo/Types/JSON.swift
```diff
@@ -28,6 +28,15 @@ public extension JSON {
...
+    switch self {
+    case let .Object(o): return o[key] ?? .Null
+    default: return .Null
```
jferris: Does this lose the type mismatch error when an unexpected JSON type is encountered? This will end up complaining that the key was missing (in `guardNull`) rather than saying that an object was expected, right?

Argo/Types/JSON.swift
tonyd256: we initially had this as a subscript `json[key]` but decided that felt weird so made this function instead ... I'd also be ok with `objectForKey` that's pretty standard for Cocoa APIs

jferris:
Is `decodeObject` part of the public API?

If it's not (and it isn't being used internally), it seems like it could be removed.

If it's public, I feel like `decodeObject` should be composed in terms of `jsonForKey`. I also feel like `jsonForKey` should probably return a `Decoded<A>` and preserve the fact that a type mismatch may occur.

tonyd256:
@jferris good stuff ... I'm going to take a look into that

Argo/Operators/Operators.swift
```diff
@@ -19,7 +19,9 @@ public func <|?<A where A: Decodable, A == A.DecodedType>(json: JSON, key: Strin
 
 // Pull embedded value from JSON
 public func <|<A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [String]) -> Decoded<A> {
-  return flatReduce(keys, json, <|) >>- A.decode
+  return keys.reduce(pure(json)) { accum, key in
```
gfontenot: Bummer that this had to go. I liked this line.

tonyd256:
@jferris taking a closer look at this, I don't think we can compose `decodedObject` in terms of `jsonForKey`. The former is used to decode the given object where as the later is used just to pull out a nested `JSON`. The `JSON` that `jsonForKey` returns can be anything, not just an `.Object`. I do still think that we need to fix `jsonForKey` to return the correct `Decoded` case for type mismatch though.

tonyd256:
@jferris and @gfontenot can you take another look at this ... the function `decodedJSONForKey` allows us to capture the type mismatch again and bring back `flatReduce` for concise operator functions.

Argo/Types/StandardTypes.swift
```diff
@@ -69,6 +69,20 @@ public func decodeObject<A where A: Decodable, A == A.DecodedType>(value: JSON)
   }
 }
 
+func decodedJSONForKey(json: JSON, key: String) -> Decoded<JSON> {
```
gfontenot: Should we explicitly mark this as `private`?

gfontenot:
One question, but it looks good to me. We still see the same performance benefits moving to the free function and using `flatReduce` again?

Argo/Types/StandardTypes.swift
tonyd256: we can't because it's being used in the `Operators.swift` file ... but it won't be visible to the public interface because it's internal by default

tonyd256:
Yes, the performance benefits are still there. It wasn't so much that any particular function call was taking too long. The problem was that we were calling some functions unnecessarily many times.

PR #134, opened 2015-05-22
Add whitespace after operator names in operator definitions.

jnutting:
Previously, all our operator definitions looked like `func ***<T>(args)` or `func ***(args)`. For many operator names, having it butt up directly against a `<` or `(` is somewhat ugly, even though it's syntactically OK. This PR changes all of those to include a space. This style has been previously established in places like ReactiveCocoa, Box, and Apple's own documentation.

### DIFF
--- a/Argo/Extensions/Dictionary.swift
+++ b/Argo/Extensions/Dictionary.swift
@@ -1,7 +1,7 @@
 import Runes
 
 // pure merge for Dictionaries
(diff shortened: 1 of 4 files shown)


kaishin:
:+1:

tonyd256:
:+1: feel free to merge into master yourself of I'll do it later today.

PR #135, opened 2015-05-26
Remove `return` from single-expression closure

alskipp:


alskipp:
As single-expression closures implicitly return their result, I've removed a couple of redundant return keywords.

tonyd256:
merged in 82696d9de5c3221b2a51cbf59b8979c41d4eabe1

Thanks for the PR!

Issue #131, opened 2015-05-11
Converting Strings to different Types during decoding


lorlab:
Hi! Thanks for this great library! I have a question about transforming values while decoding. Someone has asked the same question before but I still don't understand... And the RoleType example in the docs in not clear to me... Also I read that things might have changed in 1.0 so I thought I'd ask the same question again. Sorry if I missed the explanation somewhere.

Say I have a basic Photo model like below (this doesn't compile), how would I convert a String in the JSON to a NSURL, or a String into a NSDate with a certain format? 

```
struct Photo {
    let id: Int
    let url: NSURL
    let date: NSDate
}

extension Photo: Decodable {
    
    static func create(id: Int)(url: NSURL)(date: NSDate) -> Photo {
        return Photo(id: id, url: url, date: date)
    }
    
    static func decode(j: JSON) -> Decoded<Photo> {
        return Photo.create
            <^> j <| "id_photo"
            <*> j <| "url"
            <*> j <| "date"
    }
}
```

MichaelMcGuire:
There are multiple ways to do this, but what I've ended up doing was adding `Decoable` protocol compliance to existing types using extensions.  Here is an example for `NSDate` that converts from a `yyyy-MM-dd` format using the Objective-C runtime to hold an instance of an `NSDateFormatter` alongside the class to avoid costly creation for each decode:

```
import Foundation
import ObjectiveC
import Argo

private var jsonFormatterAssociatedKey: UInt8 = 0

extension NSDate: Decodable {
    
    class var jsonFormatter: NSDateFormatter {
        
        if let jsonFormatter = objc_getAssociatedObject(self, &jsonFormatterAssociatedKey) as? NSDateFormatter {
            return jsonFormatter
        } else {
            let jsonFormatter = NSDateFormatter()
            jsonFormatter.dateFormat = "yyyy-MM-dd"
            objc_setAssociatedObject(self, &jsonFormatterAssociatedKey, jsonFormatter, objc_AssociationPolicy(OBJC_ASSOCIATION_RETAIN))
            return jsonFormatter
        }
    }
    
    public static func decode(j: JSON) -> Decoded<NSDate> {
        switch j {
        case let .String(s):
            if let date = self.jsonFormatter.dateFromString(s) {
                return pure(date)
            } else {
                return .TypeMismatch("\(j) is not a date")
            }
        default: return .TypeMismatch("\(j) is not a date")
        }
    }
}
```

siemensikkema:
[...]

PR #140, opened 2015-06-03
Playground example update

alskipp:
The current version of Argo includes a Playground file, which I think is a great idea to demonstrate usage of the library. I've added a JSON file into the Playground `Resources` folder and shown an example of decoding the file into a `User` struct.

Anyone wanting to test out Argo’s functionality can now simply edit the example JSON file, or add new JSON files to the Playground.

I've also added an additional Playground example that shows how to decode a more complicated JSON file. It demonstrates the use of a helper function to convert to NSDate (thanks to #131). 

* * *
It is often the case when decoding JSON that the bits we're interested in are a few levels deep inside the structure. In the new Playground file, I've shown a simple technique to extract only the relevant level of JSON before decoding into a struct. I'd be interested in hearing whether you think this is a reasonable approach, or whether you'd recommend a different method?

An example of what I mean:

```swift
typealias JSONDict = [String: AnyObject]

func toJSONArray(key:String)(_ j:JSONDict) -> [JSONDict]? {
  return j[key] as? [JSONDict]
}

let appInfoFromJSONfile = { file in JSONFromFile(file) >>- toJSONArray("results") >>- first }
let app: App? = appInfoFromJSONfile("tropos") >>- decode
```

I've tried to keep the Playgrounds as short as possible while still being useful examples. Let me know what you think, I'm happy to make any changes you think are needed.


Playgrounds/iTunes-Example.playground/Resources/tropos.json
```diff
@@ -0,0 +1,9 @@
...
+              "releaseNotes":"- Fixed an issue where the app could crash if you're going buck-wild messing around with the pull to refresh control.\n- Improved the relative weather description in cases where it's crazy hot or ridiculously cold outside. Tropos will no longer tell you that it's hotter today when it's 10°F outside. (Sorry, Boston)\n- Fixed an issue where the high and low temperatures could be lower or higher than the current temperature, respectively. That's just silly.\n\nFinally, we want to thank everyone for supporting Tropos!", "minimumOsVersion":"8.0", "formattedPrice":"$0.99", "averageUserRating":4.5, "userRatingCount":10}]
+}
```
tonyd256: since we have control of this file let's put a new line at the end

Playgrounds/Argo-Basics.playground/Resources/user_with_email.json
```diff
@@ -0,0 +1,5 @@
+{
+  "id": 1,
+  "name": "Cool User",
+  "email": "u.cool@example.com"
+}
```
tonyd256: new line here as well

Playgrounds/Argo-Basics.playground/Contents.swift
```diff
@@ -0,0 +1,49 @@
...
+The **User** struct has three properties, one of which is an Optional value.
+
+(The example JSON file can be found in the **Resources** folder.)
+*/
+struct User: Printable {
+  let id: Int
+  let name: String
+  let email: String?
+  
+  var description: String {
```
tonyd256: What about moving `Printable` to an extension so the `User` struct stays clean?

Playgrounds/iTunes-Example.playground/Contents.swift
```diff
@@ -0,0 +1,76 @@
...
+
+An example JSON file (**tropos.json**) can be found in the **resources** folder.
+*/
+struct App: Printable {
+  let name: String
+  let formattedPrice: String
+  let averageUserRating: Float?
+  let releaseDate: NSDate
+  
+  var description: String {
```
tonyd256: same here, re: separate extension

Playgrounds/iTunes-Example.playground/Resources/tropos.json
tonyd256: also, pass this whole file through a good json formatter. in vim `gg = G` would be fine.

Playgrounds/iTunes-Example.playground/Contents.swift
```diff
@@ -0,0 +1,76 @@
...
+*/
+typealias JSONDict = [String: AnyObject]
+
+func JSONFromFile(file: String) -> JSONDict? {
+  return NSBundle.mainBundle().pathForResource(file, ofType: "json")
+    >>- { NSData(contentsOfFile: $0) }
+    >>- { NSJSONSerialization.JSONObjectWithData($0, options: NSJSONReadingOptions(0), error: nil) as? JSONDict }
+}
+
+func toJSONArray(key:String)(_ j:JSONDict) -> [JSONDict]? {
```
tonyd256: I don't think we need this.

Playgrounds/iTunes-Example.playground/Contents.swift
```diff
@@ -0,0 +1,76 @@
...
+      <*> j <|? "averageUserRating"
+      <*> (j <| "releaseDate" >>- toNSDate)
+  }
+}
+/*:
+When decoding JSON, the data that we're interested in is often nested a level or two inside the JSON structure.
+The iTunes JSON format contains a "results" Array and we're only interested in the first entry.
+The function below extracts the nested JSON structure containing the app info.
+*/
+let appInfoFromJSONfile = { file in JSONFromFile(file) >>- toJSONArray("results") >>- first }
```
tonyd256: I think we should be able to use `{ $0 <|| "results" }` here ... let me try it out and get back to you

alskipp:
Cheers for the feedback. I've moved `Printable` into an extension and formatted the JSON.

Playgrounds/iTunes-Example.playground/Contents.swift
tonyd256: you can remove this function completely

Playgrounds/iTunes-Example.playground/Contents.swift
```diff
@@ -0,0 +1,78 @@
...
+/*:
+* * *
+*/
+let app: App? = appInfoFromJSONfile("tropos") >>- decode
```
tonyd256: use this instead: `let app: App? = JSONFromFile("tropos")?["results"] >>- decode >>- first`

Playgrounds/iTunes-Example.playground/Contents.swift
tonyd256: remove this as well

alskipp:
Thanks for your input @tonyd256, I was stumped on how best to extract only the JSON required…
```swift
let app: App? = JSONFromFile("tropos")?["results"] >>- decode >>- first
```
… was just what I was missing. Cheers 👍

I think I've made all the suggested amends now. Let me know if any more are needed.

tonyd256:
Related to that last portion, check out this issue and feel free to leave your thoughts on the matter. #141 

Otherwise, I think I'm ok with this in general. I'd like to have @gfontenot look over it too and give his thoughts before a merge.

Playgrounds/Argo-Basics.playground/Contents.swift
```diff
@@ -0,0 +1,53 @@
...
+import Foundation
+import Argo
+import Runes
+/*:
+**Helper function** – load JSON from a file
+*/
+func JSONFromFile(file: String) -> AnyObject? {
+  return NSBundle.mainBundle().pathForResource(file, ofType: "json")
+    >>- { NSData(contentsOfFile: $0) }
+    >>- { NSJSONSerialization.JSONObjectWithData($0, options: NSJSONReadingOptions(0), error: nil) }
```
gfontenot: lets pass `nil` instead of `NSJSONReadongOptions(0)` here.

Playgrounds/iTunes-Example.playground/Contents.swift
```diff
@@ -0,0 +1,66 @@
...
+import Foundation
+import Argo
+import Runes
+/*:
+**Helper function** – load JSON from a file
+*/
+func JSONFromFile(file: String) -> AnyObject? {
+  return NSBundle.mainBundle().pathForResource(file, ofType: "json")
+    >>- { NSData(contentsOfFile: $0) }
+    >>- { NSJSONSerialization.JSONObjectWithData($0, options: NSJSONReadingOptions(0), error: nil) }
```
gfontenot: lets pass `nil` here as well

gfontenot:
I dig it.

tonyd256:
@alskipp Thanks for this! Merged in 9d55345286f44c88a887083e18cef65627934fd6

alskipp:
Cheers folks!

PR #147, opened 2015-06-13
Swift 2.0 - Remove Box from .gitmodules and minor cleanup related to Box, update Runes 

sync:
Box submodule entry was still present inside .gitmodules cause the following error

```
*** Checking out Argo at "651568ba61b58fa9112746a887e2887afd39a205"
Parse error: expected submodule commit SHA in ls-tree output: 
```

also cleaned up Box references form README and updated podspec (untested)
and update Runes to use latest Swift 2.0 branch in order to avoid the following error msg:

```
<unknown>:0: error: module file was created by an older version of the compiler: /Users/anthonymittaz/Projects/Marketplacer/YauShop/Carthage/Build/iOS/Runes.framework/Modules/Runes.swiftmodule/x86_64.swiftmodule
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "Carthage/Checkouts/Box"]
-	path = Carthage/Checkouts/Box
-	url = https://github.com/robrix/Box.git
 [submodule "Carthage/Checkouts/Runes"]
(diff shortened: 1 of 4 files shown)


tonyd256:
This is great, was running up against this issue myself. Thanks!

PR #153, opened 2015-06-15
Completely exchange Decoded for throws

tonyd256:
Ok ok ... hear me out ... it's not completely awful.

Removing the dependency on a `Result` like type means we don't really have a "monad" in here anymore (even though `throws` is basically a monad with the context being error or not). This means we can remove all the function programming stuff that some people were weary of. Also, no more `create` functions.

However, things like `map` and `reduce` got all messed up so I had to use some `for-in` statements.

Also, then on the outside to users of Argo, they will have to use the `do-try-catch` syntax so they cannot easily ignore errors but maybe that's the point the Swift team was trying for.

Anyways, I see lots of pluses and minuses for this, but I'd like to see what others think.


Argo/Extensions/Array.swift
```diff
@@ -0,0 +1,9 @@
+extension Array {
+  func mapThrows<T>(@noescape transform: Generator.Element throws -> T) throws -> [T] {
```
tonyd256: I named this differently because naming it `map` overshadows the actual `map` (the one without `throws`) ... this has got to be a bug

Argo/Operators/Operators.swift
```diff
@@ -8,44 +6,72 @@ infix operator <||? { associativity left precedence 150 }
...
 
 // Pull optional value from JSON
-public func <|? <A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<A?> {
-  return .optional(json <| key)
+public func <|? <A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) throws -> A? {
+  do {
+    return try .Some(json <| key)
+  } catch let error as DecodedError {
+    switch error {
+    case .MissingKey(_): return .None
```
tonyd256: this is now the way we need to check to see if a missing key error is "OK" for the case where the user says it may or may not be there

Argo/Types/JSON.swift
```diff
@@ -16,7 +13,7 @@ public extension JSON {
 
     case let v as [Swift.String: AnyObject]:
       return .Object(v.reduce([:]) { (var accum, elem) in
-        let parsedValue = (self.parse <^> elem.1) ?? .Null
+        let parsedValue = self.parse(elem.1) ?? .Null
```
tonyd256: this never really needed to be `map` for some reason

gfontenot: does it need the nil coalescer?

tonyd256: nope .. good catch

Argo/Types/Decodable.swift
```diff
@@ -1,4 +1,4 @@
 public protocol Decodable {
   typealias DecodedType = Self
-  static func decode(json: JSON) -> Decoded<DecodedType>
+  static func decode(json: JSON) throws -> DecodedType
```
felixjendrusch: ~~`typealias DecodedType = Self`~~ and `static func decode(json: JSON) throws -> Self`?

ikesyo: I think that cannot be used for non-final class types.

Argo/Types/DecodedError.swift
```diff
@@ -0,0 +1,13 @@
+public enum DecodedError: ErrorType {
+  case TypeMismatch(String)
+  case MissingKey(String)
+}
+
+extension DecodedError: CustomStringConvertible {
```
felixjendrusch: Not sure this is really necessary: `String(DecodedError.MissingKey("id"))` should result in `"DecodedError.MissingKey("id")"`.

Argo/Types/DecodedError.swift
```diff
@@ -0,0 +1,13 @@
+public enum DecodedError: ErrorType {
```
felixjendrusch: Have you thought about naming this just `Error`? It's still `Argo.Error` then.

Argo/Types/Decodable.swift
felixjendrusch: Ah, right, too bad :/

gfontenot: Yep, that's exactly right. This typealias exists to help with :ghost: INHERITANCE :ghost:

gfontenot:
I don't know how I feel about this, to be honest. Few general thoughts:

1. The `<|` family of operators don't really have a place in this implementation anymore. Those were useful for improving readability in an applicative style function, but here they are just noisy. We could replace those with `decode` instance methods or something.
2. Is this the same lib? It doesn't look/act/feel like Argo 1.0. It's arguably "more Swift-like", but I'm still not sure that it's a net positive change.

tonyd256:
Definitely agree about the operators not being really needed anymore.

And same with 2. I don't think this is the same library. Definitely not what we envisioned for Argo, but if it's better ... 

I am not sure it is better. I think I'll see how swift 2.0 plays out first.

MarcoSero:
Just my 2 cents here. I agree this is probably not what you guys envisioned for Argo, which at the beginning was probably experimenting with some cool functional concepts in Swift.
But this does make the library easier to use to the average Swift developer.

I also like the create is no longer necessary.

gfontenot:
> But this does make the library easier to use to the average Swift developer.

I think this is where we need to decide what we want Argo to be. IMO, we don't _need_ this to be the most popular JSON lib, and we don't _need_ this to be the most accessible JSON lib. I'm OK with Argo being "more complex" if that added complexity also makes it more powerful. In this case, I think that basing Argo around Applicative makes the library infinitely more powerful than removing that and using `throws` as laid out in this PR. For example, on our current project we used the partial application to make side loaded JSON objects dead simple. That would have been much more difficult using a non-applicative solution. Also, there are already libs that have taken the concepts in Argo and re-implemented them in a non-functional environment.

> I also like the create is no longer necessary.

fwiw, this will be true in Swift 2.0 regardless. Our recommendation for Swift 2.0 will be to curry the `init` function for your objects.

piercifani:
Since I just shipped the project I've been working on I feel I should weight in as well in this very interesting discussion.

I've built a relatively simple app, from top to bottom, using [Swift](https://developer.apple.com/swift/), [Alamofire](https://github.com/Alamofire/Alamofire), [Result](https://github.com/antitypical/Result),  [Deferred](https://github.com/bignerdranch/Deferred/) and, more importantly, Argo.

Argo has helped me a lot during development time, saving me tons of hours of debugging while developing both the app and the backend, since Argo makes obvious what's going wrong. Conversations like "Man, you're giving me an `Int` instead of a `String` in the `id` field" is way better than: "WTF? why has the app stopped working since you deployed the new backend version?". And it saved me from writing a lot boilerplate code on the viewController level because I no longer needed to check for nil or bogus values in my model layer.

This has been possible since Argo's design goals stated in their [ideology](https://github.com/thoughtbot/Argo#ideology) sections is spot on. And more: it would have been very, very difficult to achieve this in a non functional way in ObjC.

However, working with Argo with non-trivial model objects does have it's penalties: the compiler goes nuts indexing and producing binaries to test on device or simulator. Editing the `decode` function makes `SourceKit` flip out and if you make a mistake (using the wrong `<|` operator, missing a parameter in the `create` function, or other small errors) you'll end up with hours wasted scratching your head trying to find out what's going on. And sometimes it's not even possible, as noted by #106 or #5.

This penalty is so big, so big, that when reaching the deadline I preferred buying a maxed out Macbook Pro rather than making it work better on my Macbook Air.

Which brings me to my next point: it's so hard to understand what's going on behind the `decode` function that I'd prefer fork down hard earned cash for a brand new Mac rather than get cranking and contribute to this library with code rather than opening an issue. It's almost impossible for a Cocoa developer to help you guys making this awesome library more awesome.

To wrap up: I think that making Argo feel more comfortable in Swift 2.0 and it's compiler is paramount for Swift developers **without** compromising the goals stated in the ideology. Everything else (custom operators, functors, etc) are just implementation details, cool ones, but just details. 

Again, I can't thank you enough for the hard work you've put in here, and regarding this PR feel free to take whatever decision you think it's the best for this software project.

All the best!

mokagio:
Just add my humble opinion to the many ones here. What makes Argo stand out is the fact that it pushes the boundaries of how Swift can be used in a function way, see the ["Functional Concepts"](https://github.com/thoughtbot/Argo#functional-concepts) section of the README. The library name itself is a clear statement of this functional imprint.

Getting rid of Argo's functional style that would mean being incoherent with the _core values_ of the library, so to speak.

Mind me, there would be nothing bad in having a JSON parsing library that uses a more conventional, at least in the face of Swift 2, approach. But if you really want to go to that path, you should consider forking Argo into another library, because it wouldn't be Argo anymore.

I like looking at Argo, and other of your projects like [Runes](https://github.com/thoughtbot/Runes) and [Curry](https://github.com/thoughtbot/Curry), as a _functional vanguard_ for Swift. Projects that people interested in those concept can use to start getting their hands dirty, and use them as a trampoline to jump to _real_ functional languages.

After all I've seen tons of JSON parsing library, but only one Argo.

tonyd256:
This was a fun experiment but I'll going to close this as Argo isn't going to go down this route. However, there will be helper functions, something like `dematerialize` and `materialize` that will help to convert `Decoded` to a `throws` style error.

PR #157, opened 2015-06-22
Add a nil coalescing operator

terhechte:
This pull requests implements a nil coalescing operator as well as a test case for it. The use case for this operator is any JSON result where a value can be in multiple places due to historical API cruft. One example is the Instagram API, where the outer data envelope can look either like this (simplified):

```{"meta": {"code": 200}...}```

or like this:

```{"code": 200...}```

The second version only happens for specific error responses

Use it as follows:
```<*> (j <|? ["meta", "code"]) ?? (j <| "code")```

The operator requires the left hand side to be of the type `Decoded<T?>` (i.e. use `<|?`) otherwise it cannot safely work if the left hand side is nil. Also, one of the two sides needs to be non-nil, so that the expression evaluates to `T` and not `T?`.

### CODE REVIEW

**Comment by tonyd256** [comment]
So I see there really being 2 type signatures of `??`: `(A?, A) -> A` and `(A?, A?) -> A?`. The former being simply a default kind of exchange which I think we can implement with our `Decoded<A>` type like so: `Decoded<A>.value ?? A`; however, the latter is harder to perform. I would suggest that the `??` operator for `Decoded` to look more like this: `(Decoded<A>, Decoded<A>) -> Decoded<A>`

Does that make sense?

**Comment by jferris** [comment]
I think you want a different operator for `??` for `Decoded` if you're going to return a decoded value. The types don't line up:

```
(Optional<A>, A) -> A
(Decoded<A>, Decoded<A>) -> Decoded<A>
```

The analog to this in Haskell is `<|>` from the [`Alternative` typeclass](https://hackage.haskell.org/package/base-4.8.0.0/docs/Control-Applicative.html#v:-60--124--62-).

You can define both `??` and `<|>` meaningfully and following the same type structure for `Optional` and `Decoded`:

```
?? : (Optional<A>, A) -> A
?? : (Decoded<A>, A) -> A

<|> : (Optional<A>, Optional<A>) -> Optional<A>
<|> : (Decoded<A>, Decoded<A>) -> Decoded<A>
```

(I used `Optional<A>` instead of `A?` because I think it makes it easier to see the types in this case)

**Comment by tonyd256** [comment]
I actually prefer Joe's advice over my own. I think we could implement an `alternative` function on `Decoded` for now that has that `<|>` type signature. Then possibly consider adding `<|>` to Runes.

**Comment by terhechte** [comment]
I concur that a <|> operator would be more useful in this case. In addition to the points Joe already mentioned, it would also not be constrained to the 2 possible values case, but could instead even be used for use cases where there're 3 candidates (though I doubt that this is often the case).

Would it also make sense to have the types defined so that <|> could be used for alternative default values?
i.e.
`<|> : (Decoded<A>, A)`

`j <| "code" <|> j <| ["meta", "code"] <|> 200`

**Comment by tonyd256** [comment]
For the default value, I think `??` should be implemented.

**Comment by terhechte** [comment]
Right, now that I think about it, `??`  is the better solution for default values as the types match up.

**Comment by terhechte** [comment]
Hey, I've added a `<|>` operator to `decoded` and modified the test accordingly. I've defined it with a lower precedence, so it can easily be combined with the other operators to form longer expressions without having to use parentheses. I.e.

`<*> j <| "possible-key" <|> j <| "possible-key2"`

File: ArgoTests/JSON/user_with_nested_name.json [hunk 1]
```diff
@@ -0,0 +1,5 @@
+{
+  "id": 1,
+  "userinfo": {"name": "Very Cool User"},
+  "email": "u.cool@example.com",
+}
```
tonyd256: new line here

Argo/Operators/Operators.swift
```diff
@@ -49,3 +50,10 @@ public func <||? <A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [S
   return .optional(json <|| keys)
 }
 
+// MARK: Alternative operator
+public func <|><A>(lhs: Decoded<A>, rhs: Decoded<A>) -> Decoded<A> {
+  if case .Success(_) = lhs {
```
tonyd256: cool new use of the swift 2 stuff

tonyd256:
If you add that new line and rebase/squash against `td-swift-2` I'll merge it in

terhechte:
Thanks! I've updated it with a new line for the inner dictionary.

ArgoTests/JSON/user_with_nested_name.json
```diff
@@ -0,0 +1,7 @@
+{
+  "id": 1,
+  "userinfo": {
+        "name": "Very Cool User"
+      },
+  "email": "u.cool@example.com"
+}
```
tonyd256: still need a new line at the end of this file

Argo/Operators/Operators.swift
tonyd256: can this be just `if case .Success = lhs {`?

terhechte:
Sorry for that. I had a look at the other json files, and those didn't have newlines either, so I thought you meant the indentation of the inner dictionary. Just pushed with a newline. Also, used the simplified  `case .Success = lhs` form that you mentioned. Interestingly I wouldn't have thought that that works, but indeed it does, pretty neat.

tonyd256:
Great thanks! Ya, I probably need to fix those JSON files. I'll squash and merge today

terhechte:
Awesome, thanks!

tonyd256:
Thanks again for the PR! Merged in here: 5b18ce0da13e3e6d8a3a5188cbf00bd8d5c86a0c

Issue #161, opened 2015-06-29
Is there an equivalent to a <|?| operator?


alcarvalho:
What I mean by that is getting back an array of optionals, i.e. `[User?]`.

Or any other way of getting an array of a type, but ignore any items in which we receive errors?

alcarvalho:
It won't make it. As discussed on #164.

PR #164, opened 2015-07-04
A resilient list decoding operator: <|?|

alcarvalho:
Thinking about how to sove #161 I have come up with this operator.

I think that just omitting items that could not be decoded makes more sense then returning an array of optionals `[T?]`.

I am not sure about how I have named things though.

### DIFF
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -7,6 +7,10 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		58A6EB6EF0673E74EB6B4215 /* ResilientPosts.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58A6ECC0D7FC701226390B68 /* ResilientPosts.swift */; };
+		58A6EE00114C7A89B93933A8 /* ListOfOptionalsDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58A6EED845300864E8BD46E0 /* ListOfOptionalsDecodingTests.swift */; };
+		6579A3681B482055008D8451 /* ResilientPosts.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58A6ECC0D7FC701226390B68 /* ResilientPosts.swift */; };
+		6579A3691B482077008D8451 /* ListOfOptionalsDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58A6EED845300864E8BD46E0 /* ListOfOptionalsDecodingTests.swift */; };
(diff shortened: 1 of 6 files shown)


tonyd256:
@alcarvalho thanks for making this PR; however, I'd like to stray away from adding more custom operators to Argo. As it is now, there is a lot of resistance to use Argo because of the custom operators and adding one more won't help. That being said, I think you can accomplish what you want without an operator.

First, are you sure that you actually want an array of optionals? This PR doesn't seem to be adding any functionality. It appears that you've added functions with the same type signature as functions that were already in Argo. Even their implementation is the same. Also, the example you use in this PR is not an array of optional `Comment`s but just an array of non-optional `Comment`s which the `<||` operator would be fine for.

Anyways, to accomplish the `[Comment?]` that you want I would do this:
```swift
switch json {
case let .Array(a): return a.map { Comment.decode($0).value }
default: return []
}
```

gfontenot:
Note that another way to accomplish this would be to use a custom parser that you `flatMap` into the pipeline:

```swift
j <|| "comments" >>- { pure($0.map { Comment.decode($0).value }) }
```

(types might be slightly off, but that's the general idea.)

alcarvalho:
@tonyd256 ok, I understand what you're saying. I am sure you want to stay away from more complexity.

That being said, I have changed my mind about the array of optionals, because I thought that just ignoring (filtering) the objects that did not decode from the array in the first place would be even better, as I stated in the first comment.

This is why the operator has the exact same signature as the `<||` because it produces the same array type, but ignoring errors of each value decoded. That's also what I meant by "resilient". ;)

But I am totally ok if you think that this is not a good thing to be added.

@gfontenot thank you for that line. That will be helpful in case you don't want to accept this implementation. :)

gfontenot:
> I have changed my mind about the array of optionals, because I thought that just ignoring (filtering) the objects that did not decode from the array in the first place would be even better, as I stated in the first comment.

FWIW: This can be done quickly by throwing a [`concatOptionals`](https://github.com/thoughtbot/ExtendedArray) into the mix:

```swift
j <|| "comments" >>- { pure(concatOptionals($0.map { Comment.decode($0).value })) }
```

Also, note that this can easily be generalized to a small custom parser function:

```swift
func decodeFilteredArray<T: Decodable where T == T.DecodedType>(js: [JSON]) -> Decoded<[T]> {
  return pure(concatOptionals(js.map { T.decode($0).value }))
}
```

That'll work for any array of objects that are decodable.

alcarvalho:
Thanks again, @gfontenot! 

So it seems that you guys think this is a bad idea, so I am cloning this PR. ;)

gfontenot:
@alcarvalho Just to be clear, we don't think this is a bad idea at all. We just want to be careful about keeping the main lib focused. For things like this, we prefer educating users on how to build their own solutions that work _with_ Argo.

alcarvalho:
That totally makes sense! You have been very supportive to me. Thanks! ;)

Issue #166, opened 2015-07-10
Is there a way to parse two dimensional arrays?


lfrahm:
Hey there!

First of all thanks for providing such a convenient way of transforming JSON data into models! It is super helpful. 

But I'm wondering how I can parse more-dimensional arrays. I can't just wrap it into a new model since the deeper JSON text doesn't have a field, and unfortunately there is no <||| operator.

So is there a way to parse something like:
```{
    "data":[
        [0,1],
        [2,3],
        [4,5]
    ]
}```

**Comment by lfrahm** [comment]
So here is what I can come up with. I would appreciate some feedback.

For file Operators.swift
```swift
infix operator <||| { associativity left precedence 150 }
infix operator <|||? { associativity left precedence 150 }

// MARK: Arrays of Arrays

// Pull array of arrau from JSON
public func <||| <A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<[[A]]> {
    return json <| key >>- decodeArrayOfArray
}

// Pull optional array of array from JSON
public func <|||? <A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<[[A]]?> {
    return .optional(json <||| key)
}

// Pull embedded array of array from JSON
public func <||| <A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [String]) -> Decoded<[[A]]> {
    return json <| keys >>- decodeArrayOfArray
}

// Pull embedded optional array of array from JSON
public func <|||? <A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [String]) -> Decoded<[[A]]?> {
    return .optional(json <||| keys)
}

```

For file file StandardTypes.swift 
```swift
public func decodeArrayOfArray<A where A: Decodable, A == A.DecodedType>(value: JSON) -> Decoded<[[A]]> {
  switch value {
    case let .Array(a):
      var twoD: [[Decoded<A>]] = []
      for elem in a {
        switch elem {
        case let .Array(b): twoD.append(A.decode <^> b)
        default: return typeMismatch("Array", value)
        }
      }
      return sequence(twoD)
    default: return typeMismatch("Array", value)
  }
}
```
and for file sequence.swift
```swift
func sequence<T>(xs: [[Decoded<T>]]) -> Decoded<[[T]]> {
    return reduce(xs, pure([])) { accum, elem in
        curry(+) <^> accum <*> (pure <^> sequence(elem))
    }
}
```
[...]

PR #170, opened 2015-07-15
Make sequence public

tonyd256:
This is a really useful function for decoding more complex types like
multi-dimensional arrays. We should open it up to users of Argo.


tonyd256:
This is really useful for something like #166

gfontenot:
I'm :+1: on this. Didn't realize it wasn't already public.

PR #171, opened 2015-07-15
Remove Coaching from README

gfontenot:
We don't actually offer this anymore, so it's probably best to not
advertise it.

/cc @croaky @jferris


tonyd256:
:+1:

PR #172, opened 2015-07-15
Remove reference to Box.framework from project

gfontenot:
This was accidentally left in, even though it was removed from the
source and wasn't linked anymore.


tonyd256:
oh huh .. how did that sneak through ... :+1:

PR #174, opened 2015-07-15
Remove Runes as an external dependency

gfontenot:
This removes Runes as an external dependency, preferring to use it as an
internal dependency that only serves to define the operators for map,
apply, and flatMap.

The benefit here is that for users of Argo, they can simply import Argo
in their models instead of needing to import Argo _and_ Runes. But at
the same time, we can still rely on Runes to provide a centralized place
for the definition of the map, apply, and flatMap operators. This means
Argo will still play nice with Runes, or other libs that use the same
version of Runes to define these operators.

Internally, we weren't using much of Runes anyway, and the parts we
_were_ using can be implemented as instance methods instead. Note,
especially, how simple the transition from `>>-` to `.flatMap` is.

### DIFF
--- a/Argo Playground.playground/Pages/Basics.xcplaygroundpage/Contents.swift
+++ b/Argo Playground.playground/Pages/Basics.xcplaygroundpage/Contents.swift
@@ -5,15 +5,14 @@
 */
 import Foundation
 import Argo
-import Runes
 import Curry
 /*:
 **Helper function** – load JSON from a file
 */
 func JSONFromFile(file: String) -> AnyObject? {
   return NSBundle.mainBundle().pathForResource(file, ofType: "json")
-    >>- { NSData(contentsOfFile: $0) }
-    >>- JSONObjectWithData
+    .flatMap { NSData(contentsOfFile: $0) }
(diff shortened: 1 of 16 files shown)


jshier:
Just when I was starting to get used to `>>-`. 

I like not having to import Runes anymore, since it makes my manual dependency management that much easier. :+1:  :disappointed_relieved:

gfontenot:
Don't worry, you'll still have `>>-` available. But it will only be implemented for `Decoded<T>`, and won't be available for `Optional<T>` or `Array<T>`. For those, you'd need to explicitly declare Runes as a dependency. Same goes for `<^>`, `<*>`, and `pure`.

tonyd256:
:+1: This is great! We should do something in the README though about Runes. It is useful for some trickier JSON decodes.

gfontenot:
@tonyd256 Do you think I should do that in this PR? Or should we roll that into the larger documentation initiative that we're still totally planning on doing and hasn't been put on the back burner at all?

tonyd256:
Ya the latter. I think that will be important when we talk about decoding more complex models in the docs.

PR #175, opened 2015-07-16
Use Curry.framework instead of local curry

gfontenot:
We really shouldn't be the ones supplying `curry` for use with Argo.
Instead, we should encourage the use of a centralized dependency. This
will allow users to use the version of `curry` that they _want_ to use,
and will reduce the risk of collisions.

Since this isn't a hard dependency for the framework, I'm keeping it
private. Users can choose to ignore it altogether and write a curried
`create` function the same way they used to if they want. But the
suggested method will be to use `curry`, and we'll recommend
[`Curry.framework`][curry] to fit that purpose.

[curry]: https://github.com/thoughtbot/Curry


tonyd256:
:+1: I like it

PR #176, opened 2015-07-16
Remove Runes references from tests + cleanup

gfontenot:
Accidentally missed some references to Runes in the tests (no clue how
this compiled with those references) so those are gone now.

Also taking the opportunity to clean up our use of init when currying. I
personally prefer using `self.init` instead of specifying the model
name. I'd _rather_ just do `curry(init)`, but the compiler isn't super
smart about that yet and complains about us trying to define an
initializer.


tonyd256:
:+1:

PR #178, opened 2015-07-20
fix missing Curry Xcode project

sync:
if using `thoughtbot/Curry` without a branch and there is a tagged release version with a binary attached on github, Carthage will only fetch the precompiled framework and won't fetch the associated Xcode project required by Argo

### DIFF
--- a/Argo.xcworkspace/xcshareddata/Argo.xcscmblueprint
+++ b/Argo.xcworkspace/xcshareddata/Argo.xcscmblueprint
@@ -4,13 +4,14 @@
 
   },
   "DVTSourceControlWorkspaceBlueprintWorkingCopyStatesKey" : {
-    "57C931977B7D2307CC013C2BD93F90CF7C676790" : 0,
+    "194A7A54F3CE0138A6C320E0643B7C60F8218891" : 0,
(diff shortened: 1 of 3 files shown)


Argo.xcworkspace/xcshareddata/Argo.xcscmblueprint
```diff
@@ -4,13 +4,14 @@
 
   },
   "DVTSourceControlWorkspaceBlueprintWorkingCopyStatesKey" : {
-    "57C931977B7D2307CC013C2BD93F90CF7C676790" : 0,
+    "194A7A54F3CE0138A6C320E0643B7C60F8218891" : 0,
```
gfontenot: This file shouldn't be checked in at all. I'm planning on removing it at some point.

gfontenot:
I don't think this is the proper solution for this. We should be advising people to run `bin/setup`, which will use the proper flags with Carthage to ensure that we checkout the project instead of downloading the binaries.

sync:
i completely missed the `bin/setup`, so I guess this PR is no longer needed

sync:
just tried `bin/setup` it's working, very well. added issue #179

Issue #114, opened 2015-04-24
Handling [String:String] dictionaries


edwardaux:
Hi, I've been using Argo in a new Swift project and am loving it so far - it is my first real foray into functional programming so pretty much everything is a learning curve :-)

I've managed to create my own model objects and can successfully parse my JSON except for one last thing - handling `[String:String]` data. Imagine that my (simplified) model object looks something like this:

	struct User {
		let name: String
		let age: Int
		let metadata: [String:String]
	}
	
My JSON would look like this:

	{
		"name" : "John Smith",
		"age"  : 42,
		"metadata" : {
			"field1" : "some value",
			"otherField" : "foo bar"
		}
	}
	
The `metadata` property is really just a collection of random name/value pairs. There could be none, or there could be 50 and I don't know ahead of time what the keys are (hence it is modelled as a `[String:String]`)

Unfortunately, `Decoded<[String:String]>` is not a supported return type for the `<|` operator. At the moment, my half-baked solution is to create an object that represents this type of dictionary and then traverse over the `JSON` values (see below). But it just doesn't feel right. Just curious if there is a better way?  Any thoughts or guidance would be much appreciated.  Thanks.

	struct StringDictionary {
		let values: [String:String]
		init(_ values: [String:String]) {
			self.values = values
		}
	}
	extension StringDictionary : JSONDecodable {
		static func decode(j: JSON) -> Decoded<StringDictionary> {
			switch j {
			case .Object(let s):
				var values = [String:String]()
				// TODO not sure how to traverse a list of un-named keys
				return .Success(Box(StringDictionary(values)))
			default: return .TypeMismatch("\(j) is not a valid [String:String] dictionary")
			}
		}
	}

gfontenot:
This is actually something that we (accidentally) implemented in the SOON TO BE RELEASED WE PROMISE Argo 1.0. Decoding to dictionaries (of they type `[String: T: JSONDecodable]`) Just Works™ like any other decodable type.

edwardaux:
Thanks for the quick reply @gfontenot. Is that code in a private branch, or is something I can merge to my fork in the meantime?

gfontenot:
The change is in `master` now. It's also a significant change from `0.4.3` in terms of API. You might want to look over the diff in #112 if you want to get a sense of the changes. The code on `master` is _probably_ what we'll be shipping for 1.0, although I can't make that guarantee, and likewise can't make any guarantee about any non-official release.

Take that as you will.

[...]

PR #181, opened 2015-07-27
Expose functions outside module to support easier extension

edwardaux:
A little while back I asked about how to handle `[String:String]` - see #114. Thanks to some changes in Argo and your guidance in that issue, I could create my own operators and life was good.

In the move to Swift 2, though, some of the requisite functions have had their visibility reduced so aren't accessible when defining operators outside the Argo module.

For example, the following custom operators to support `[String:Decodable]` barf because `flatReduce` and `decodedJSON` aren't visible. I started by copy/pasting those functions, but I also needed to copy/paste `guardNull` and `typeMismatch` and it started to get ugly.

	public func <|~ <A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<[String:A]> {
		return json <|~ [key]
	}
	public func <|~? <A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<[String:A]?> {
		return .optional(json <|~ [key])
	}
	public func <|~ <A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [String]) -> Decoded<[String:A]> {
		return flatReduce(keys, initial: json, combine: decodedJSON) >>- Dictionary<String,A>.decode
	}
	public func <|~? <A where A: Decodable, A == A.DecodedType>(json: JSON, keys: [String]) -> Decoded<[String:A]?> {
		return .optional(json <|~ keys)
	}

This PR stops that boilerplate copying by defining `flatReduce` and `decodedJSON` as public.

### DIFF
--- a/Argo/Functions/flatReduce.swift
+++ b/Argo/Functions/flatReduce.swift
@@ -1,4 +1,4 @@
-func flatReduce<S: SequenceType, U>(sequence: S, initial: U, combine: (U, S.Generator.Element) -> Decoded<U>) -> Decoded<U> {
+public func flatReduce<S: SequenceType, U>(sequence: S, initial: U, combine: (U, S.Generator.Element) -> Decoded<U>) -> Decoded<U> {
   return sequence.reduce(pure(initial)) { accum, x in
     accum >>- { combine($0, x) }
   }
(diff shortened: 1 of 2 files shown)


tonyd256:
I don't see why not. :+1: I can merge this later or if @gfontenot has time today maybe he can get to it.

gfontenot:
:dancers:

PR #200, opened 2015-08-27
Simplify object parsing.

paulyoung:
This pull request simplifies the conversion from `[Swift.String: AnyObject]` to it's JSON equivalent.

The use of the nil coalescing operator in the existing implementation seems somewhat redundant since `.Null` is already the default case for the `parse` method.

Perhaps I'm missing something here so would love to understand if I got something wrong.

### DIFF
--- a/Argo/Types/JSON.swift
+++ b/Argo/Types/JSON.swift
@@ -1,5 +1,4 @@
 import Foundation
-import Runes
 
 public enum JSON {
   case Object([Swift.String: JSON])
@@ -13,7 +12,7 @@ public extension JSON {
   static func parse(json: AnyObject) -> JSON {
     switch json {
     case let v as [AnyObject]: return .Array(v.map(parse))
-    case let v as [Swift.String: AnyObject]: return .Object(parse <^> v)
+    case let v as [Swift.String: AnyObject]: return .Object(v.map(parse))
     case let v as Swift.String: return .String(v)
     case let v as NSNumber: return .Number(v)
     default: return .Null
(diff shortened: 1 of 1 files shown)


Argo/Types/JSON.swift
```diff
@@ -16,8 +16,7 @@ public extension JSON {
 
     case let v as [Swift.String: AnyObject]:
       return .Object(reduce(v, [:]) { (var accum, elem) in
-        let parsedValue = (self.parse <^> elem.1) ?? .Null
-        accum[elem.0] = parsedValue
+        accum[elem.0] = self.parse(elem.1)
```
paulyoung: `self` can be inferred here if preferred.

Argo/Types/JSON.swift
```diff
@@ -13,14 +13,7 @@ public extension JSON {
...
     case let v as [AnyObject]: return .Array(v.map(parse))
-
-    case let v as [Swift.String: AnyObject]:
-      return .Object(reduce(v, [:]) { (var accum, elem) in
-        let parsedValue = (self.parse <^> elem.1) ?? .Null
-        accum[elem.0] = parsedValue
-        return accum
-      })
-
+    case let v as [Swift.String: AnyObject]: return .Object(parse <^> v)
```
paulyoung: I'd like to suggest that the line above should also use the map operator.

Otherwise, I can remove my use of it for consistency which would also mean that this file no longer depends on Runes.

paulyoung: Although Runes doesn't define a map operator for `Dictionary`, removing the import at this point causes a build error with the message `Operator is not a known binary operator`.

tonyd256:
@paulyoung This is great thanks! We started down this route in the Swift 2.0 branch, but for some unknown reason didn't go as far as you've taken it here. Would you mind making this change to the Swift 2.0 branch instead since we'll probably be merging that into master in a few weeks 🙏

paulyoung:
@tonyd256 done in #201 :+1:

PR #204, opened 2015-08-30
[Proposal] Default implementation for RawRepresentable

swizzlr:
With this commit, you can now do this:

```swift
enum SomeRemoteDataTypeStatus: String {
  case Success = "success"
  case Failure = "failed"
  case Indeterminate = "ok" // legacy API
}

extension SomeRemoteDataTypeStatus: Decodable {}
```

Commit Message:
```
This commit introduces extensions to Decodable providing default implementations of `decode` for certain `RawRepresentable` types.
    
For instance, an enum backed by a String can now be implicitly decoded by Argo simply by declaring that it is `Decodable`.
    
An implementation using `NSNumber`, which is the most generally applicable type, is introduced too. We might want to copy paste a ton more if we like this approach.
```

As an aside, let me ask if on the basis of the following, what is even the point of Argo or am I just in a daze:

```swift
public extension Decodable where Self.DecodedType == Self, Self: RawRepresentable, Self.RawValue == [String:JSON] {
  private typealias Objectish = Self.DecodedType
  public static func decode(json: JSON) -> Decoded<Self> {
    switch json {
    case .Object(let object):
      let res = self.init(rawValue: object)
      let dec = Decoded<Objectish>.fromOptional(res)
      return dec
    default:
      return Decoded.TypeMismatch("Attempted to decode \(json) into \(self), but needed an Object")
    }
  }
}

struct DictionaryRepresentedStruct: RawRepresentable {
  typealias RawValue = [String:JSON]
  let foo: Int
  init?(rawValue: [String:JSON]) {
    if let fooJSON = rawValue["foo"], case .Number(let fooValue) = fooJSON {
      foo = fooValue as Int
    } else {
      foo = 0
      return nil
    }
  }
  var rawValue: [String:JSON] {
    return ["foo":JSON.Number(foo)]
  }
}

extension DictionaryRepresentedStruct: Decodable { }
```

Obvious downside of coupling type implementation to JSON aside (which could be fixed with multiply protocol extensions for `[String:String]` and other types of values), it seems that RawRepresentable is the standard library's big daddy of "make a type more general, suitable for use in data encoding; decode a type from data" and might make a good foundation for future work with Argo, like so:

```
Using Argo

Implement `RawRepresentable`; call `YourType.decode`; have a beer.
```


swizzlr:
Duh; the downside to general `init?rawValue` implementations is that you lose the error information from the `Decoded` monad. The upside is that if JoSchmo comes up with `my awesome persistence framework that uses rawrepresentable dot swift` then you're good to go

(all hail the standard library for it is mighty in its slicing)

paulyoung:
I imagine instead of a new file, we might want this to be added to https://github.com/thoughtbot/Argo/blob/td-swift-2/Argo/Types/StandardTypes.swift

paulyoung:
In line with my previous comment, the implementations could then make use of the private `typeMismatch` function and keep the messaging consistent.

paulyoung:
Totally forgot to say, I think this is a pretty cool concept :heart:

swizzlr:
Thanks – all your comments are totally valid, but I definitely want to first know if this is even a direction Argo should take. I'm happy enough using the extension myself. Exposing public extensions to stdlib types isn't strictly a good thing.

tonyd256:
Interesting ... I'm on vacation for a couple of weeks while I move to SF so I might not get to take a deep look at this for a bit.

Argo/Extensions/RawRepresentable.swift
```diff
@@ -0,0 +1,30 @@
+public extension Decodable where Self.DecodedType == Self, Self: RawRepresentable, Self.RawValue == String {
+  // Work around beta 6 bug where types cannot be deduced properly:
+  private typealias Stringish = Self.DecodedType
+  public static func decode(json: JSON) -> Decoded<Self> {
+    switch json {
+    case .String(let string):
+      let res = self.init(rawValue: string)
+      let dec = Decoded<Stringish>.fromOptional(res)
+      return dec
```
gfontenot: instead of using the intermediary variables, we can reduce this to:

```swift
return .fromOptional(Self.init(rawValue: string))
```

Argo/Extensions/RawRepresentable.swift
```diff
@@ -0,0 +1,30 @@
+public extension Decodable where Self.DecodedType == Self, Self: RawRepresentable, Self.RawValue == String {
+  // Work around beta 6 bug where types cannot be deduced properly:
+  private typealias Stringish = Self.DecodedType
```
gfontenot: I'm not seeing this issue. I think we can remove these lines.

Argo/Extensions/RawRepresentable.swift
```diff
@@ -0,0 +1,30 @@
...
+  public static func decode(json: JSON) -> Decoded<Self> {
+    switch json {
+    case .String(let string):
```
gfontenot: can we do `case let .String(s)` to follow the style of the rest of the project?

Argo/Extensions/RawRepresentable.swift
```diff
@@ -0,0 +1,30 @@
...
+      let res = self.init(rawValue: string)
+      let dec = Decoded<Stringish>.fromOptional(res)
+      return dec
+    default:
+      return Decoded.TypeMismatch("Attempted to decode \(json) into \(self), but needed a String")
```
gfontenot: you can omit `Decoded` here and just do `.TypeMismatch`

Argo/Extensions/RawRepresentable.swift
```diff
@@ -0,0 +1,30 @@
...
+    }
+  }
+}
+
+public extension Decodable where Self.DecodedType == Self, Self: RawRepresentable, Self.RawValue == NSNumber {
+  // Work around beta 6 bug where types cannot be deduced properly:
```
gfontenot: same comments from above apply here as well

Argo/Extensions/RawRepresentable.swift
```diff
@@ -0,0 +1,30 @@
+public extension Decodable where Self.DecodedType == Self, Self: RawRepresentable, Self.RawValue == String {
+  // Work around beta 6 bug where types cannot be deduced properly:
+  private typealias Stringish = Self.DecodedType
+  public static func decode(json: JSON) -> Decoded<Self> {
```
gfontenot: can we change `json` to `j`? that follows the style for the rest of the standard types.

gfontenot:
This is awesome. I'm totally in favor of pulling this in. I think it's a fantastic change. I'm actually going to pull the string version of this in for the project I'm on now because I'm tired of writing these kinds of decoders.

swizzlr:
Thanks for saying! I'll tidy up the above style as you suggest (apologies, getting SourceKit to obey you can be quite tiring).

swizzlr:
@gfontenot shall I focus this pull request around strings and remove the number decoding for now?

gfontenot:
I think both are useful for the same reasons. I'd add both because why not. They are going to look almost identical but ¯\_(ツ)_/¯

swizzlr:
BRB writing tests after the fact

paulyoung:
Development-Driven Tests :trollface:

tonyd256:
@swizzlr This looks great! I think this is something we would like to pull in. If you could rebase against the branch again and fix it up per @gfontenot's comments then we can take another look at it!

tonyd256:
@swizzlr I'm going to merge the swift 2 branch into master tomorrow. If you can work on this today that would be great! If not, no worries. You'll have to reopen the PR against master.

tonyd256:
@swizzlr We like this and want to pull it in. If you don't have time right now that's fine. I'm going to branch off your code and finish it up. This will keep your commits.

tonyd256:
finished up in #226

tonyd256:
Thanks again @swizzlr! This was merged in as part of #226

PR #208, opened 2015-09-04
Explicitly code sign framework

sidraval:
Carthage is unhappy without this, ala https://github.com/thoughtbot/Swish/pull/3

### DIFF
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -850,6 +850,7 @@
 		D0592EBB1B77DD7800EFEF39 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				CODE_SIGN_IDENTITY = "iPhone Developer";
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				DEFINES_MODULE = YES;
 				DYLIB_COMPATIBILITY_VERSION = 1;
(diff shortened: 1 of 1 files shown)


tonyd256:
Interesting .. is this new? Also, does this cover us for OS X and watchOS?

sidraval:
Yeah, @gfontenot had problems installing Swish via Carthage, and I'm having problems installing Argo via Carthage now.

@gfontenot mentioned in Slack that it might be an XCode bug... but this seems to help for now?

I'm not sure about OSX/watchOS.

gfontenot:
We need to do this for watchOS as well, but not for OS X

sidraval:
@tonyd256 @gfontenot Any other comments / questions?

gfontenot:
@sidraval Can you make this change for the watchOS target as well?

sidraval:
@gfontenot I made the change at the `Argo` project level, not at the specific iOS target level, so does this already apply to the watchOS target?

When I select the watchOS target specifically, I see its `Code Signing Identity` set to `iOS Developer` already.

gfontenot:
Ah, I didn't pull the project down to check.

We should be setting this at the target level, not the project level. Setting it at the project level will (I believe) cause issues for the Mac target.

sidraval:
@gfontenot Alright, I made it for both the `Argo-iOS` and `Argo-watchOS` targets individually.

gfontenot:
:+1: merge away

Issue #207, opened 2015-09-03
Argo in Swift 2.0


alvinvarghese:
I integrated Argo with my project ( Swift 1.2, Xcode 6.4 ). I wanna move to the Xcode 7 beta 6, with Swift 2.0. I want to integrate argo with Swift 2.0. When i imported project to Xcode 7 beta 6, it suggested lot of code changes from Swift 1.2 to Swift 2.0. Which was really bad, i really could not understand what Xcode did to my project.

Could you please tell me that, which is the best way, updating via cocoapods ? Do i have to wait till the release of Xcode 7, Cause i decided that i will be taking Argo with my app to the production. So i don't want to waste time by integrating again and solving the bugs.

May i wait till the release of Xcode 7 ? or this Argo gonna behave the same without any problem in Xcode 7.

Suggest me your thoughts.

ikesyo:
You should check #142, and you might be able to use the branch of that PR as following: 

```ruby
pod 'Argo', :git => 'https://github.com/thoughtbot/Argo.git', :branch => 'td-swift-2'
```

alexkent:
I have an issue #203 with Cocoapods integration on the td-swift-2 branch.
I would be very interested to hear if you don't run into this!

ikesyo:
Oh, I'm sorry. It seems that we should add `:submodules => true` to the entry above.

alvinvarghese:
Okay, Let me update it, let you guys know :)

alexkent:
aha, Cocoapods ignores the `submodules => true` in the .podspec when integrating with `:git => '...'`
As discussed here https://github.com/CocoaPods/CocoaPods/issues/2317

paulyoung:
My understanding is that the dependency on Runes was removed in the `td-swift-2` branch.

alexkent:
The dependency on including the entire Runes framework was removed. But a single "Runes.swift" files is still included (and necessary as it declares the custom operators) when building the `td-swift-2` branch.

The source files line in the podspec includes it:
`spec.source_files = 'Argo/**/*.{h,swift}', 'Carthage/Checkouts/Runes/Source/Runes.swift'`

gfontenot:
yep, that's correct. We've removed Runes.framework as a dependency, but still rely on it as an internal dependency.

The snippet that @ikesyo pasted _should_ be correct. I'm going to close this for now, and if it's still an issue after using that format, please let me know. We will ship a proper pod once we've finalized support for Swift 2.0.

[...]

Issue #203, opened 2015-08-28
Cocoapods integration broken in td-swift-2 branch, Runes.swift not included


alexkent:
I'm integrating Argo in my project Swift 2.0 project via Cocoapods.
The relevant line in my podfile looks like this (to get this https://github.com/thoughtbot/Argo/commit/b2be8edc080d937bbdc6c43b29a9d1e5084c312c commit)

`pod 'Argo', :git => 'https://github.com/thoughtbot/Argo', :commit => 'b2be8edc080d937bbdc6c43b29a9d1e5084c312c'`

After pod install, the Argo framework will not build in my project because the Runes.swift file is not included. This file is explicitly included in the Argo.podspec, yet it does not get included during install by Cocoapods 0.38.2.

Any ideas?

alexkent:
The same problem (i think) can be seen by running `pod spec lint`or `pod lib lint` on the Argo repo.

alexkent:
adding `:submodules => true` in the podfile resolves this issue. as discussed here #207

PR #209, opened 2015-09-04
Removed Runes submodule.

paulyoung:
This pull request removes the old submodule entries for Runes.

I believe that Runes is no longer used on this branch and this information was left around due to https://github.com/Carthage/Carthage/issues/442.

I looked into this after reading #207.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "Carthage/Checkouts/Runes"]
-	path = Carthage/Checkouts/Runes
-	url = https://github.com/thoughtbot/Runes.git
 [submodule "Carthage/Checkouts/Curry"]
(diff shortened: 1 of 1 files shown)


paulyoung:
My mistake – the more relevant issue is #203, but #207 is related.

gfontenot:
So we removed the dependency on Runes.framework, but actually still depend on Runes internally for the operator definitions. See #174 for more info.

paulyoung:
I see, you're just including the file from the submodule. Makes sense!

gfontenot:
Yep! It's a bit out of the norm, but it means we can use the monadic operators without having to expose Runes to users. It also means that Argo will Just Work with Runes if you happen to include both.

paulyoung:
Great change.

PR #210, opened 2015-09-04
Change Decoded to look more like Result

gfontenot:
This ditches the strategy of using multiple error states, for using a
single error state with multiple possible error causes.

The end result here is a simpler Decoded type, without sacrificing the
flexibility and expressiveness that we'd run into by switching to Result
itself. Additionally, this will make transforming from Decoded to Result
completely trivial:

```swift
func makeResult<T>(d: Deocded<T>) -> Result<T, DecodeError> {
 switch d {
 case let .Success(value): return .Success(value)
 case let .Failure(error): return .Failure(error)
 }
}
```

I've also added helper functions for failing states. It was always kind of
annoying to have to create these yourself, even more so now that you have to
wrap multiple types. This adds some more semantic static functions so that it's
cleaner to create these if needed.

### DIFF
--- a/Argo/Types/Decoded.swift
+++ b/Argo/Types/Decoded.swift
@@ -1,17 +1,3 @@
-public enum DecodeError: ErrorType {
-  case TypeMismatch(expected: String, actual: String)
-  case MissingKey(String)
-}
-
-extension DecodeError: CustomStringConvertible {
-  public var description: String {
-    switch self {
-    case let .TypeMismatch(expected, actual): return "TypeMismatch(Expected \(expected), got \(actual))"
-    case let .MissingKey(s): return "MissingKey(\(s))"
-    }
-  }
-}
-
 public enum Decoded<T> {
   case Success(T)
   case Failure(DecodeError)
--- a/Argo/Types/Decoded.swift
+++ b/Argo/Types/Decoded.swift
@@ -1,12 +1,12 @@
 public enum DecodeError: ErrorType {
-  case TypeMismatch(String)
+  case TypeMismatch(expected: String, actual: String)
   case MissingKey(String)
 }
 
 extension DecodeError: CustomStringConvertible {
   public var description: String {
     switch self {
-    case let .TypeMismatch(s): return "TypeMismatch(\(s))"
+    case let .TypeMismatch(expected, actual): return "TypeMismatch(Expected \(expected), got \(actual))"
     case let .MissingKey(s): return "MissingKey(\(s))"
     }
   }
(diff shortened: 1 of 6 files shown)


gfontenot:
Note that this will _also_ make it simple to support `throws` via a `dematerialize` function, like Result does.

paulyoung:
These changes seem like a great way to acknowledge where the language appears to be heading without compromising too much.

jshier:
Since this doesn't seem to address the limitation of consumers not being able create completely custom error types, I'd still like the ability to create .TypeMismatch with a fully custom string rather than having to conform to the pattern currently provided. Or perhaps I'll have to switch to .MissingKey, even if that makes a bit less sense. Or perhaps provide a .Custom or .General state we can pass back in our custom decoding?

gfontenot:
@jsteiner I didn't realize custom error types were something people needed. Can you open an issue and we can discuss that separately?

jshier:
@gfontenot I assume you meant me. I'll open a new issue.

gfontenot:
@jshier: ha, yes. Sorry @jsteiner, you got autocompleted.

tonyd256:
I think this is a good step in the right direction. We can look more into custom error states in future issues/PRs.

Argo/Types/Decoded.swift
```diff
@@ -15,25 +14,38 @@ public extension Decoded {
...
   }
 }
 
+public extension Decoded {
+  static func typeMismatch<T, U: CustomStringConvertible>(expected: String, actual: U) -> Decoded<T> {
+    return .typeMismatch(expected, actual: "\(actual)")
+  }
+
+  static func typeMismatch<T>(expected: String, actual: String) -> Decoded<T> {
+    return .Failure(.TypeMismatch(expected: expected, actual: "\(actual)"))
```
tonyd256: we can remove the string interpolation here

Argo/Types/Decoded.swift
```diff
@@ -102,8 +111,8 @@ public func <^> <A, B>(f: A -> B, a: Decoded<A>) -> Decoded<B> {
 
 /**
   apply a `Decoded` function to a `Decoded` value
-  - If the function is a failing case (`.TypeMismatch`, `.MissingKey`), this will return the function's value
-  - If the function is a `.Success` case and the value is a failure case, this will return the value
+  - If the function is `.Failure`, this will return the function's failure info
+  - If the function is a `.Success` case and the value is `Failure`, this will return the value's failure info
```
tonyd256: `.Failure`?

tonyd256: I know this was like this before these changes but I think we lost some consistency with this
maybe have it more like this: `If the function is `.Success` and the ...`

Argo/Types/StandardTypes.swift
```diff
@@ -73,7 +73,7 @@ public extension Dictionary where Value: Decodable, Value == Value.DecodedType {
   static func decode(j: JSON) -> Decoded<[String: Value]> {
     switch j {
     case let .Object(o): return sequence(Value.decode <^> o)
-    default: return typeMismatch("Object", forObject: j)
+    default: return .typeMismatch("Object", actual: "\(j)")
```
tonyd256: same comment as above

Argo/Types/StandardTypes.swift
```diff
@@ -81,17 +81,13 @@ public extension Dictionary where Value: Decodable, Value == Value.DecodedType {
 public func decodedJSON(json: JSON, forKey key: String) -> Decoded<JSON> {
   switch json {
   case let .Object(o): return guardNull(key, j: o[key] ?? .Null)
-  default: return typeMismatch("Object", forObject: json)
+  default: return .typeMismatch("Object", actual: "\(json)")
```
tonyd256: once more for the record

tonyd256:
a few consistency comments but it's looking great!

paulyoung:
:metal:

PR #213, opened 2015-09-08
Make sequence public

angerman:
I have a case where I need to call `sequence` explicitly as I'm using custom decoders for a sum type. (each case has a separate decoder). And I'm turning `[JSON]`, into `[MySumType]`.

Does it hurt to have `sequence` public?


gfontenot:
Huh, weird. We've already done this on `master`, not sure why we haven't done it here as well. `td-swift-2` seems to be a bit out of sync with `master`.

gfontenot:
I think what we _should_ do here is rebase `td-swift-2` against `master` to make sure they are in sync. I just did a dry run and there were only a couple of conflicts, so that should work fine. I don't want to make that decision without input from @tonyd256, though, and he's traveling right now.

angerman:
Weird. Well I'll have my branch to work off of, in the mean time. Let's wait for @tonyd256 to decide.
Your rebase idea sounds like a sensible one.

tonyd256:
ya we should do the rebase ... it's about time with Swift 2 coming out. I'll take care of this.

tonyd256:
After a rebase from master this is all set. I'll close this now, but feel free to comment here if you have questions about this.

PR #214, opened 2015-09-08
Updates in the README file

AlexanderKaraberov:
Added info and workaround in the README.md about the known issue with Swift compiler:  "Expression was too complex to be solved in reasonable time; consider breaking up into distinct sub-expressions". I think this info should be in the Readme because it is quite popular.

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,6 +341,50 @@ You can decode custom types the same way, as long as the type also conforms to
 
 For more examples on how to use Argo, please check out the tests.
 
+### Known Issues
+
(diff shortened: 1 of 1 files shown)


gfontenot:
This is great. I think we _might_ end up holding onto this for now, though. We're still planning on doing a major documentation overhaul (see the `documentation` branch), and adding this to the README is going to conflict. If you want to open a PR against _that_ branch, I think we could merge it sooner.

AlexanderKaraberov:
@gfontenot  Ok. Got it. I will close this one and open a PR against `documentation` branch to avoid duplication.

PR #92, opened 2015-03-26
Write them docs

tonyd256:
# TODO:

- [x] Update syntax for Swift 2.0

## Meta
- [x] Ideology
- [x] Functional Concepts

## Basic Usage
 - [x] Using the global `decode` function
 - [x] Writing a simple parser
 - [ ] Relationships

## Advanced Usage
 - [ ] Decoding an Array of Arrays
 - [ ] Date parsing using a function
 - [ ] NSURL Parsing using a function
 - [ ] Conforming external types to `Decodable`
 - [ ] Decoding Enum types
 - [ ] Decoding objects with a root key
 - [ ] Partially decoding to support side loaded objects

PR #216, opened 2015-09-08
Update README.md

AlexanderKaraberov:
Added info and workaround for the known issue with Swift compiler: "Expression was too complex to be solved in reasonable time; consider breaking up into distinct sub-expressions". It will be reasonable to create a separate file "Known Issues" if there will be more problems which have solutions.


tonyd256:
Thanks for this! We know there is a lack of documentation for Argo and we want to tackle it at some point so these kinds of contributions are awesome! I'll review it soon and get back to you.

for reference #92

tonyd256:
Would you mind actually moving this into a separate file: `Documentation/Known-Issues.md`? Then you can reference it from `Documentation/README.md`

AlexanderKaraberov:
Of course no.
I just wanted such kind of information was accessible in some way to avoid further confusions.

tonyd256:
I need to merge in the documentations branch to keep master up to date. That will automatically close this PR but would you mind re-opening this PR on master?

PR #218, opened 2015-09-12
Also require app only extension API for watchOS

nickygerritsen:
#46 enabled this for iOS and Mac OS X, but it was still disabled for the watchOS target


tonyd256:
@nickygerritsen Thanks for this! Merged in here: 096288b78225656cba7d6008c95959d3988809a5

PR #220, opened 2015-09-15
Support optional returns of decoding operators

tonyd256:
This is a bit of duplication, but I see a lot of users' code using `.value` to ignore the `Decoded` type. Is this something people would find useful?


sidraval:
I like it! I found myself using `.value` today, in fact.

Does this remove the ability to access the `Decoded` version? Is there any use case for that one?

tonyd256:
@sidraval you can use both the `Optional` or `Decoded` versions of all those operators. As long as you type the return, type-inference will figure out which to use. Yes, the `Decoded` version has the use case of retaining information on why decoding failed.

sidraval:
Ah, I see. In that case, I'm definitely in favor :+1:

gfontenot:
This removes the need to use `value`, but I'm concerned about the added complexity. Overloading operators comes at a performance cost as well as a cognitive overhead cost.

(god i wish we had monads)

tonyd256:
So just for sanity, I ran this and looked at the 2 `measureBlock`s in the tests. There time was not significantly larger than without this. So I don't think there is a runtime performance hit. I also tried a full clean and build but also did not see a time difference between the branches.

So I think the Swift compiler is "OK" with the complexity ... at least for how the tests use it.

There is still the issue of the cognitive complexity. But I think this might be easier than to think you need to have `.value` in certain places. I think I'd like to try having this and we can see if we like it or not.

PR #221, opened 2015-09-15
Move alternative declaration to Decoded

sidraval:
@gfontenot Mentioned this probably belonged in the `Decoded` file rather than the Operators file, since it is specific to `Decoded`... thoughts?


tonyd256:
what about it's own file?

gfontenot:
I think it makes sense to move this alongside the `map`, `apply`, `flatMap` implementations. We could move it to its own thing if that makes more sense to you though. I think we all agree that it shouldn't be stuck at the bottom of Operators though

tonyd256:
I agree, @sidraval how about you take alternative and the operator definitions of map, flatMap, and apply and move them all into a new file. This will sharpen the focus of the Decoded Type file and the Argo operators file.

tonyd256:
Also, if you can make the doc comments all nice and fancy like the other operators have, that would be great.

gfontenot:
It might be worth pulling out the monadic operators into their own file as well?

tonyd256:
@gfontenot I think that's what I was suggesting ... that the implementations for `>>-`, `<^>`, `<*>`, and `<|>` should move into a new file. BUT the implementations for `map`, 'flatMap`, and `apply` can stay as extensions in the Decoded file.

sidraval:
Sounds good, will resubmit shortly

Argo/Operators/DecodedOperators.swift
```diff
@@ -0,0 +1,69 @@
...
+/**
+  flatMap a function over a `Decoded` value (right associative)
+
+  - If the value is a failing case (`.TypeMismatch`, `.MissingKey`), the function will not be evaluated and this will return the value
+  - If the value is `.Success`, the function will be applied to the unwrapped value
+
+  - parameter a: A value of type `Decoded<A>`
+  - parameter f: A transformation function from type `A` to type `Decoded<B>`
+
+  - returns: A value of type `Decoded<U>`
```
tonyd256: U should be B

Argo/Operators/DecodedOperators.swift
```diff
@@ -0,0 +1,69 @@
...
+  - parameter f: A `Decoded` transformation function from type `A` to type `B`
+  - parameter a: A value of type `Decoded<A>`
+
+  - returns: A value of type `Decoded<B>`
+*/
+public func <*> <A, B>(f: Decoded<A -> B>, a: Decoded<A>) -> Decoded<B> {
+  return a.apply(f)
+}
+
+// MARK: Alternative operator
```
tonyd256: extra empty line after this

PR #222, opened 2015-09-15
Remove custom string description

tonyd256:
This basic implementation of a custom string now comes by default when
using enums. We can remove this and the default description looks good
enough.

An example is: `Object(["id": Argo.JSON.Number(1), "email":
Argo.JSON.String("u.cool@example.com"), "name": Argo.JSON.String("Cool
User")])`

### DIFF
--- a/Argo/Types/JSON.swift
+++ b/Argo/Types/JSON.swift
@@ -30,18 +30,6 @@ extension JSON: Decodable {
   }
 }
 
-extension JSON: CustomStringConvertible {
-  public var description: Swift.String {
-    switch self {
-    case let .String(v): return "String(\(v))"
-    case let .Number(v): return "Number(\(v))"
-    case let .Array(a): return "Array(\(a.description))"
-    case let .Object(o): return "Object(\(o.description))"
-    case .Null: return "Null"
-    }
-  }
-}
-
 extension JSON: Equatable { }
 
 public func == (lhs: JSON, rhs: JSON) -> Bool {


jshier:
Why remove this? While the new generated description may be "good enough", the current version looks much better, as it removes quite a bit of redundant info, especially for large models. Part of my current project's logging is to log the intermediary Argo JSON representation of the responses we get back. Changing it to this would increase the length of those logs by a fair bit. I suppose I could add my own JSON extension, but I just don't see the downside to keeping this in the project.

tonyd256:
@jshier No real downside to keep it in if it's useful. I didn't know people were using it in a way beyond debugging. I'll like to hear from others about this but it may not get merged in the end.

tonyd256:
An alternative: change the description to output the data as if it were just json and not show the types ... ie. `["id": 1, "email": "u.cool@example.com", "name": "Cool User"]`

jshier:
Heh, I actually use this in the logging because it makes it easy to tell what the actual returned types are, since the documentation I have isn't always correct, and printing what's returned from `NSJSONSerialization` isn't always clear. Given that such information is one of the reasons for the Argo JSON type in the first place, it makes sense to me to keep it in the `description`. 

In any case, like I said, I can always add the old description code to my project if this lands.

gfontenot:
I don't really care one way or another, but I think that's a sign that it's not worth messing with. Seems to be working as is, and we don't really gain much by removing it.

sidraval:
Agree w/ @gfontenot

tonyd256:
Cool, I'm fine with that. :hocho: :gun:

PR #225, opened 2015-09-17
Add dematerialize

paulyoung:
This pull request aims to enable going from `Decoded` to `throws` and vice-versa.

Currently this doesn't compile since the `catch` for `materialize` is not exhaustive and I'm not too sure what to do there.

### DIFF
--- a/Argo/Types/Decoded.swift
+++ b/Argo/Types/Decoded.swift
@@ -76,3 +76,12 @@ public extension Decoded {
 public func pure<A>(a: A) -> Decoded<A> {
   return .Success(a)
 }
+
+public extension Decoded {
+  func dematerialize() throws -> T {
+    switch self {
+    case let .Success(value): return value
+    case let .Failure(error): throw error
+    }
+  }
+}
--- a/ArgoTests/Tests/DecodedTests.swift
+++ b/ArgoTests/Tests/DecodedTests.swift
@@ -28,4 +28,45 @@ class DecodedTests: XCTestCase {
     default: XCTFail("Unexpected Case Occurred")
     }
   }
+  
+  func testDecodedDematerializeSuccess() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_email")!)
+    
+    do {
+      try user.dematerialize()
+      XCTAssert(true)
+    } catch {
+      XCTFail("Unexpected Error Occurred")
+    }
+  }
+  
+  func testDecodedDematerializeTypeMismatch() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
+    
+    do {
+      try user.dematerialize()
+      XCTFail("Unexpected Success")
+    } catch DecodeError.TypeMismatch {
+      XCTAssert(true)
+    } catch DecodeError.MissingKey {
+      XCTFail("Unexpected Error Occurred")
+    } catch {
+      XCTFail("Unexpected Error Occurred")
+    }
+  }
+  
+  func testDecodedDematerializeMissingKey() {
+    let user: Decoded<User> = decode(JSONFromFile("user_without_key")!)
+    
+    do {
+      try user.dematerialize()
+      XCTFail("Unexpected Success")
+    } catch DecodeError.MissingKey {
+      XCTAssert(true)
+    } catch DecodeError.TypeMismatch {
+      XCTFail("Unexpected Error Occurred")
+    } catch {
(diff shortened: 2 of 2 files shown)


gfontenot:
Ha. You beat @tonyd256 to the PR. He was working on `dematerialize` last night.

I'm :+1: on `dematerialize`, but I'm not sure what the use case would be for `materialize`. I think that makes sense in Result, but I don't know if it applies here. Did you have ideas on when that would be useful?

Also, I think @tonyd256 was working on a test around `dematerialize`. Not sure if he got that working or not.

paulyoung:
@gfontenot I think I agree with you on `materialize` but thought it might be nice to cover the edge case.

I only really added it because it was mentioned here: https://github.com/thoughtbot/Argo/pull/153#issuecomment-140873022

Tests for `dematerialize` on the way, unless @tonyd256 beats me to it.

paulyoung:
Added some tests, please let me know if there is any feedback.

tonyd256:
Damn you @paulyoung!! You beat me! haha ... anyways, This looks fine except I think you're going to run into the same issue I'm currently having.

Try changing you `catch` to catch the specific error the `try` should be throwing. I can't get that to work. ... Also feel free to look at my branch for what I mean.

Argo/Types/Decoded.swift
```diff
@@ -76,3 +76,12 @@ public extension Decoded {
 public func pure<A>(a: A) -> Decoded<A> {
   return .Success(a)
 }
+
+public extension Decoded {
+  public func dematerialize() throws -> T {
```
tonyd256: don't think you need public twice here

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -28,4 +28,26 @@ class DecodedTests: XCTestCase {
     default: XCTFail("Unexpected Case Occurred")
     }
   }
+  
+  func testDecodedDematerializeSuccess() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_email")!)
+    
+    do {
+      try user.dematerialize()
+      XCTAssertTrue(true, "Materialize threw an error with Decoded.Success")
```
tonyd256: these error messages look off?

paulyoung:
@tonyd256 just pushed some changes based on your comments (and your branch).

tonyd256:
did the tests pass for you?

I finally got it to pass by adding `Argo.` to the front of those cases ... see updates to my branch

paulyoung:
@tonyd256 everything seems to work as expected. Can you confirm?

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -28,4 +28,45 @@ class DecodedTests: XCTestCase {
...
+    }
+  }
+  
+  func testDecodedDematerializeTypeMismatch() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
+    
+    do {
+      try user.dematerialize()
+      XCTFail("Unexpected Success")
+    } catch DecodeError.TypeMismatch(_, _) {
```
tonyd256: looks like we can just leave off all the `(_)` and `(_,_)`s in all these

paulyoung: Done.

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -28,4 +28,45 @@ class DecodedTests: XCTestCase {
...
+  }
+  
+  func testDecodedDematerializeTypeMismatch() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
+    
+    do {
+      try user.dematerialize()
+      XCTFail("Unexpected Success")
+    } catch DecodeError.TypeMismatch {
+      XCTAssertTrue(true)
```
tonyd256: what to shorten this to `XCTAssert(true)` and the one below

tonyd256:
@paulyoung on small change and then could you squash this down to one commit?

ArgoTests/Tests/DecodedTests.swift
paulyoung: Thanks. Thought I'd got all of those.

paulyoung:
Squashed :eggplant:

tonyd256:
:boom: :muscle: :us: .... Thanks @paulyoung

PR #227, opened 2015-09-17
Add Relationship documentation

tonyd256:


Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+      <^> j <| "author"
+      <*> j <| ["author", "name"]
+      <*> j <| "text"
+  }
+}
+```
+
+Using an array of strings allows us to traverse embedded objects to get at the
+value we want.
+
```
gfontenot: Extra newline.

gfontenot:
One comment, but I think this is great. Maybe ping @thoughtbot/copy-editing for prose feedback?

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
+It's very common to have custom models that relate to other custom models.
```
gabebw: `# Relationships` header?

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+Our `Post` model will be very simple:
+
+```swift
+struct Post {
+  let author: String
+  let text: String
+}
+```
+
+Then, implementing `Decodable` for `Post` we get this:
```
gabebw: "Our implementation of `Decodable` for `Post` looks like this:"

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+```
+{
+  "author": "Gob Bluth",
+  "text": "I've made a huge mistake.."
+}
+```
+
+Great! Now we can decode json into `Post` models. Let's be real, we can't have
+posts without comments! Comments are like 90% of the fun on the internet. (See
+reddit, YouTube, etc)
```
gabebw: I'd remove the parenthetical, leave some mystery for people who haven't been on the internet before like this guy: http://www.salon.com/2013/03/14/after_25_years_in_jail_the_internet_blew_my_mind_partner/

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+With the embedded comments array, the json may look like this:
+
+```
+{
+  "author": "Lindsay",
+  "text": "I have the afternoon free.",
+  "comments": [
+    {
+      "author": "Lucille",
+      "text": "Really? Did \"nothing\" cancel?"
```
gabebw: Instead of escaped double quotes, how about single quotes?

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+    }
+  ]
+}
+```
+
+Next, let's say the author's name isn't good enough and we want the `User`
+relationship to the `Post` and `Comment`. If you use the `User` code from
+[Basic Usage], we can simply replace the `String` type with `User` on the
+`author` property of out models.
```
gabebw: out -> our

gabebw:
I'm not qualified to review the code samples, btw.

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+    "name": "Lindsay"
+  },
+  "text": "I have the afternoon free.",
+  "comments": [
+    {
+      "author": {
+        "id": 1,
+        "name": "Lucille"
+      },
+      "text": "Really? Did \"nothing\" cancel?"
```
gabebw: Same here - single instead of escaped double qutoes

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+    }
+  ]
+}
+```
+
+Yep that's right, the only thing that changed was `String` to `User`!
```
gabebw: "was the type of `author`: it was a `String` and now it's a nested `User` hash."

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+
+We can also create a convenience property to get directly to the user's name
```
gabebw: "to directly access the user's name"

Documentation/Relationships.md
mcmire: I agree, and I'd wrap "like" in commas ("Comments are, like, ...") or remove it.

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+  static func decode(j: JSON) -> Decoded<Post> {
+    return curry(self.init)
+      <^> j <| "author"
+      <*> j <| "text"
+      <*> j <|| "comments"
+  }
+}
+```
+
+We added the array as a property on out `Post` model. Then we added a line to
```
mcmire: out -> our

mcmire: This paragraph sounds a little choppy and perhaps a little redundant. What do you think about:

> Since `comments` is an array, notice how we're using `<||` instead of `<|` to decode them.

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+extension Post: Decodable {
+  static func decode(j: JSON) -> Decoded<Post> {
+    return curry(self.init)
+      <^> j <| "author"
+      <*> j <| "text"
+  }
+}
+```
+
+And the json looks like this:
```
mcmire: json -> JSON

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
+It's very common to have custom models that relate to other custom models.
+When all your models conform to `Decodable`, Argo makes it really easy to
+populate those relationships. Let's look at a `Post` and `Comment` model and
+how they relate.
+
+Our `Post` model will be very simple:
```
mcmire: What do you think about using the present tense here? ("will be" -> "is")

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+  "author": "Gob Bluth",
+  "text": "I've made a huge mistake.."
+}
+```
+
+Great! Now we can decode json into `Post` models. Let's be real, we can't have
+posts without comments! Comments are like 90% of the fun on the internet. (See
+reddit, YouTube, etc)
+
+Let's look at a simple `Comment` model:
```
mcmire: What do you think about "Let's add a simple `Comment` model"?

Documentation/Relationships.md
mcmire: Actually, what do you think about

> Let's start with a simple `Post` model:

Documentation/Relationships.md
mcmire: Also, since this is an example, what do you think about "Encoding a Post could then produce the following JSON:"?

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+      <*> j <| "text"
+      <*> j <|| "comments"
+  }
+}
+```
+
+We added the array as a property on out `Post` model. Then we added a line to
+decode the comments array from the json. We use `<||` because it is an _Array_.
+
+With the embedded comments array, the json may look like this:
```
mcmire: may -> could

json -> JSON

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+  "comments": [
+    {
+      "author": "Lucille",
+      "text": "Really? Did \"nothing\" cancel?"
+    }
+  ]
+}
+```
+
+Next, let's say the author's name isn't good enough and we want the `User`
```
mcmire: What do you think about:

> Storing the name of the author with a post or comment isn't very flexible. What we really want to do is tie posts and comments to users. If we use the `User` class from [Basic Usage](Basic-Usage.md), we can simply change the `author` property on `Post` so that it's a `User` instead of a `String`.

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+extension Comment: Decodable {
+  static func decode(j: JSON) -> Decoded<Comment> {
+    return curry(self.init)
+      <^> j <| "author"
+      <*> j <| "text"
+  }
+}
+```
+
+Where the json now has an embedded `User`:
```
mcmire: Now the JSON for a post could look like this:

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+      "text": "Really? Did \"nothing\" cancel?"
+    }
+  ]
+}
+```
+
+Yep that's right, the only thing that changed was `String` to `User`!
+
+We can also create a convenience property to get directly to the user's name
+instead of having to deal with the model later.
```
mcmire: I don't know what "deal with the model later" means; can you reword this?

Documentation/Relationships.md
```diff
@@ -0,0 +1,193 @@
...
+  static func decode(j: JSON) -> Decoded<Comment> {
+    return curry(self.init)
+      <^> j <| "author"
+      <*> j <| ["author", "name"]
+      <*> j <| "text"
+  }
+}
+```
+
+Using an array of strings allows us to traverse embedded objects to get at the
```
mcmire: I'm assuming you're referring to this line above:

```
<*> j <| ["author", "name"]
```

Does this relate somehow to the new `authorName` property added? If so, that seems super magical to me.

Documentation/Relationships.md
mcmire: take that back, I think your wording might be better here.

tonyd256:
Thanks for the feedback everyone! This PR will be closed because I need to merge in the documentation branch so master is up to date. I'll reopen with this feedback taken in.

PR #228, opened 2015-09-17
Add decoding from root keys doc

tonyd256:


Documentation/Decode-Root-Keys.md
```diff
@@ -0,0 +1,26 @@
...
+```
+{
+  "user": {
+    "id": 1881372911,
+    "name": "George Sr"
+  }
+}
+```
+
+In this case, you can't use the global `decode` function. The easiest way
```
gfontenot: This paragraph reads a little funky to me.

gfontenot:
Imma ping @thoughtbot/copy-editing here as well

Documentation/Decode-Root-Keys.md
```diff
@@ -0,0 +1,26 @@
+The easiest way to decode models from json is to use the global `decode`
```
gabebw: I'd add a header so people have context even if they're directly linked here. How about `# Decoding Root Keys`?

Documentation/Decode-Root-Keys.md
```diff
@@ -0,0 +1,26 @@
...
+like so:
+
+```swift
+let json = JSON.parse(anyObject)
+
+let user: Decoded<User> = json <| "user"
+// or
+let user: User? = json <| "user"
+```
+
```
gabebw: extra line

Documentation/Decode-Root-Keys.md
```diff
@@ -0,0 +1,26 @@
+The easiest way to decode models from json is to use the global `decode`
+function. You can pass the `AnyObject` variable returned from
+`NSJSONSerialization` directly to `decode` and get the model back. However,
+many times the model json is embedded within a root key:
+
+```
+{
+  "user": {
+    "id": 1881372911,
+    "name": "George Sr"
```
gabebw: Sr -> Senior

Documentation/Decode-Root-Keys.md
mcmire: json -> JSON

mcmire: Also, what about: 

> The easiest way to decode models from JSON is to use the global `decode` function: pass the `AnyObject` variable returned from `NSJSONSerialization` directly to `decode`, and you'll get the model back.

Documentation/Decode-Root-Keys.md
```diff
@@ -0,0 +1,26 @@
+The easiest way to decode models from json is to use the global `decode`
+function. You can pass the `AnyObject` variable returned from
+`NSJSONSerialization` directly to `decode` and get the model back. However,
+many times the model json is embedded within a root key:
```
mcmire: model json -> JSON for the model?

Documentation/Decode-Root-Keys.md
mcmire: The easiest way what?

You've already said "the easiest way" at the beginning of this document, so what about:

> In this case, you can't use the global `decode` function [edit: is there a specific reason why?]. To get around this, first parse the `AnyObject` into `JSON`, then use the `<|` operator to ... [what are you doing here?]

tonyd256:
Thanks for the feedback everyone! This PR will be closed because I need to merge in the documentation branch so master is up to date. I'll reopen with this feedback taken in.

PR #229, opened 2015-09-18
Update Curry and Runes

tonyd256:


Cartfile.private
```diff
@@ -1,2 +1,2 @@
 github "thoughtbot/Curry"
-github "thoughtbot/Runes" "v3.0.0-rc.2"
+github "thoughtbot/Runes"
```
gfontenot: we actually _do_ need to make sure they are using `>= 3.0.0` here, I think.

PR #233, opened 2015-09-21
Add decoding from root keys doc

tonyd256:


tonyd256:
@gabebw @mcmire @gfontenot Here is the updated PR with your comments taken into account. Please give it another look when you have time.

tonyd256:
oh and @thoughtbot/copy-editing

Documentation/Decode-Root-Keys.md
```diff
@@ -0,0 +1,28 @@
+# Decoding with Root Keys
+
+The easiest way to decode models from JSON is to use the global `decode`
+function: pass the `AnyObject` variable returned from `NSJSONSerialization`
+directly to `decode`, and get the model back. However, many times the JSON for
```
mcmire: No comma necessary here.

However, what if you say this (in which case you *would* need a comma):

"pass the `AnyObject` variable ..., and you'll get the model back"

Documentation/Decode-Root-Keys.md
```diff
@@ -0,0 +1,28 @@
...
+  "user": {
+    "id": 1881372911,
+    "name": "George Senior"
+  }
+}
+```
+
+In this case, you can't use the global `decode` function because it assumes the
+object you're trying to decode is at the root level. To get around this, first
+parse the `AnyObject` into a `JSON` type then use the `<|` operator to pull out
```
mcmire: What about a comma before "then" to break things up a bit?

mcmire:
Looks good to me (other than the two comments above).

gfontenot:
:+1:

PR #234, opened 2015-09-21
Add Relationship documentation

tonyd256:


tonyd256:
Adding back in @thoughtbot/copy-editing ... this PR is based of the one that was closed and taking into account your comments. Please take another look over when you have time.

Documentation/Relationships.md
```diff
@@ -0,0 +1,195 @@
+# Decoding Relationships
+
+It's very common to have custom models that relate to other custom models.
```
gabebw: Do we need the "custom"s here? Are there predefined models anywhere?

Documentation/Relationships.md
```diff
@@ -0,0 +1,195 @@
...
+  }
+}
+```
+
+And the JSON looks like this:
+
+```
+{
+  "author": "Gob Bluth",
+  "text": "I've made a huge mistake.."
```
gabebw: One period or three, please. I'd prefer one.

Documentation/Relationships.md
```diff
@@ -0,0 +1,194 @@
...
+  static func decode(j: JSON) -> Decoded<Post> {
+    return curry(self.init)
+      <^> j <| "author"
+      <*> j <| "text"
+      <*> j <|| "comments"
+  }
+}
+```
+
+We added the array as a property on our `Post` model. Then we added a line to
```
mcmire: I think the only issue I have with this para is that you're using "array" a lot -- is there any way you can reduce that?

Documentation/Relationships.md
```diff
@@ -0,0 +1,194 @@
...
+        "id": 1,
+        "name": "Lucille"
+      },
+      "text": "Really? Did 'nothing' cancel?"
+    }
+  ]
+}
+```
+
+Yep that's right, the only thing that changed was the type of `author`: it was
```
mcmire: Here you're repeating the end of the previous paragraph. I think having a light conversational style is fine here, so what if you removed this paragraph completely, and changed the previous one to read: "we can simply change the `author` property from `String` to `User` (yup, really!)"?

Documentation/Relationships.md
tonyd256: how about:
>We added the comments as a property on our `Post` model. Then we added a line to decode the comments from JSON. Notice ....

Documentation/Relationships.md
```diff
@@ -0,0 +1,194 @@
...
+  static func decode(j: JSON) -> Decoded<Comment> {
+    return curry(self.init)
+      <^> j <| "author"
+      <*> j <| ["author", "name"]
+      <*> j <| "text"
+  }
+}
+```
+
+Using an array of strings allows us to traverse embedded objects to get at the
```
mcmire: Can you qualify where you're using an array of strings? I can see how someone who's new might have to hunt for a bit to find that what you mean is the `<*> j <| ["author", "name"]` line.

Documentation/Relationships.md
mcmire: Yeah, that's better.

Documentation/Relationships.md
```diff
@@ -0,0 +1,194 @@
...
+Our `Post` model will be very simple:
+
+```swift
+struct Post {
+  let author: String
+  let text: String
+}
+```
+
+Then, our implemention of `Decodable` for `Post` looks like this:
```
mcmire: implemention -> implementation

mcmire:
Nice! This looks better.

Documentation/Relationships.md
```diff
@@ -0,0 +1,191 @@
...
+      "text": "Really? Did 'nothing' cancel?"
+    }
+  ]
+}
+```
+
+Storing the name of the author with a post or comment isn't very flexible.
+What we really want to do is tie posts and comments to users. If we use the
+`User` struct from [Basic Usage], we can simply change the `author` property
+from `String` to `User`. No joke! Take a look:
```
gfontenot: might be good to put the `User` struct in this code sample as well.

Documentation/Relationships.md
```diff
@@ -0,0 +1,191 @@
...
+        "id": 1,
+        "name": "Lucille"
+      },
+      "text": "Really? Did 'nothing' cancel?"
+    }
+  ]
+}
+```
+
+We can also create a convenience property to directly access the user's name
```
gfontenot: this seems unrelated to relationships.

gfontenot:
Couple of small comments.

gfontenot:
Should we talk about _why_ this works? Or should that be left for an internals document?

tonyd256:
@gfontenot Perhaps we could have a short blurb like: 
"This works because any type that conforms to `Decodable` can be decoded using the Argo operators." 

Then link to the internals document like: 
"To learn more about how the Argo operators work for anything that conforms to `Decodable`, [check this out]."

dazmuda:
Hmm. Why do we go in the order 1)model 2)decodable 3)JSON?

I feel like if you're consuming JSON, you're going 1)JSON 2)model 3)decodable

As in: I read the documentation, I see JSON that looks like what I'm consuming, I read how to set up the model and make it decodable.

Just a thought.

tonyd256:
@dazmuda that's fair. I agree. When I was writing it I initially didn't put in the JSON and so when I was done throwing it in the top didn't fit right. I will try to rearrange.

gfontenot:
@tonyd256 can we get this cleaned up/into master? It's getting pretty stale.

tonyd256:
@thoughtbot/ios @thoughtbot/copy-editing re-review please!

Documentation/Relationships.md
```diff
@@ -0,0 +1,173 @@
...
+      <*> j <| "text"
+  }
+}
+```
+
+That's it! "How does this work?", you ask? Well, Argo is smart enough to know it can decode
+anything that conforms to `Decodable` because internally, Argo is simply calling
+each type's `decode` function. In this example, `Post`, `Comment`, and `User`
+all conform to `Decodable` so Argo looks at those types the same way it looks at
+`String` or `Int`.
```
gfontenot: This feels like it might make a good intro. "Argo can do this cool thing because it's generic! Here's what it looks like! See how that helped?!"

Documentation/Relationships.md
```diff
@@ -0,0 +1,173 @@
...
+        "id": 1,
+        "name": "Lucille"
+      },
+      "text": "Really? Did 'nothing' cancel?"
+    }
+  ]
+}
+```
+
+Then, we can add the `User` model to `Post` and `Comment`:
```
gfontenot: I feel like this section could be massaged a bit. I kinda had trouble keeping up with everything that was changing. Maybe you should re-introduce `User`, then show the updated JSON, then show how it can be added to `Post` and `Comment`?

I also had this thought that it's a _little_ weird to change the type of the `author` key this late into the game. What do you think about using the embedded syntax to grab the author name, then changing it to store the whole author?

tonyd256: good idea!

Documentation/Relationships.md
tonyd256: Ok I'll move it up and see how it feels.

tonyd256:
@gfontenot is this better?

dazmuda:
Oooh, I really like the change you made re Gordon's comment about keeping the author JSON consistent.

One little thing: Maybe we could store the user ID rather than the name? Since just storing a ref ID to represent a relationship is a valid use case, whereas just storing a name isn't.

Not necessary though. This looks really good. The step order is really great now.

Documentation/Relationships.md
```diff
@@ -0,0 +1,165 @@
...
+Our `Post` model could then be:
+
+```swift
+struct Post {
+  let author: String
+  let text: String
+}
+```
+
+And, our implementation of `Decodable` for `Post` looks like this:
```
gfontenot: Maybe make a note here or earlier that we only care about the user's name at this point, so we'll use the syntax for grabbing the embedded name object.

Documentation/Relationships.md
```diff
@@ -0,0 +1,165 @@
...
+extension User: Decodable {
+  static func decode(j: JSON) -> Decoded<User> {
+    return curry(User.init)
+      <^> j <| "id"
+      <*> j <| "name"
+  }
+}
+```
+
+We can simply change the `author` property from `String` to `User`. No joke!
```
gfontenot: maybe add "and point directly at the `author` key"

Documentation/Relationships.md
```diff
@@ -0,0 +1,165 @@
...
+extension Comment: Decodable {
+  static func decode(j: JSON) -> Decoded<Comment> {
+    return curry(self.init)
+      <^> j <| "author"
+      <*> j <| "text"
+  }
+}
+```
+
+That's it!
```
gfontenot: What do you think about wrapping up here with something mirroring the intro?

> Argo's use of polymorphism through generics makes supporting custom types straightforward, since you're using the exact same plumbing that Argo itself uses internally

or something like that

gfontenot:
This reads really nicely. Just a couple of small comments.

Documentation/Relationships.md
tonyd256: good call

tonyd256:
@gfontenot another final look please

gfontenot:
✨ ❤️ ✨

PR #236, opened 2015-09-21
Add Decoding Enums documentation

tonyd256:


Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,59 @@
+# Decoding Enums
+
+Decoding Structs and Classes is a similar process, but Enums can mix things up.
+If your Enum inherrits from a Raw Type `String` or `Int`, then making it conform
```
paulyoung: inherrits -> inherits

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,59 @@
...
+      }
+    default: return .typeMismatch("String", actual: j)
+    }
+  }
+}
+```
+
+The first thing we do is check to see if the `JSON` is a number. If so, we look
+for the matching enum case. Finally, we return the matching case using the
+`pure` function which will wrap the enum value in a `Decoded` type.
```
sidraval: I might add these as comments inside the code block? I found myself wanting them, and then realizing you had added exactly what I was wondering underneath.

sidraval:
Small comment, but otherwise looks good!

Documentation/Decode-Enums.md
tonyd256: do you think having both would be too much. Where the comment ones are abbreviated for what's to follow.

sidraval: IMO try and do inline ones, and if it feels like not enough go for it with additional explanation under

sidraval:
wzytmfx-compressor

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,65 @@
...
+      case 13.1: return pure(.HalfMarathon) // 13.1 miles is a half marathon distance.
+      case 26.2: return pure(.Marathon) // 26.2 miles is a marathon distance.
+      case _ where n > 26.2: return pure(.UltraMarathon) // Any distance over a marathon is considered an ultra marathon.
+
+      // Return an error if no case matched
+      default: return .typeMismatch("marathon distance", actual: n)
+      }
+
+    // Return an error if JSON is not a number.
+    default: return .typeMismatch("String", actual: j)
```
gfontenot: s/String/Number

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,65 @@
...
+
+```swift
+enum Role: String {
+  case User 
+  case Admin
+  case SuperAdmin
+}
+```
+
+To make `Role` conform to `Decodable`, use this one line:
```
gfontenot: Maybe add a reminder that in Swift 2, enums with a raw type get that type from the case name? Or maybe give this enum explicit raw values to be clearer?

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,69 @@
+# Decoding Enums
+
+Decoding Structs and Classes is a similar process, but Enums can mix things up.
```
gfontenot: s/is/are

pretty sure, anyway

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,69 @@
...
+extension Role: Decodable { }
+```
+
+"THAT'S IT?! How?", you ask. Enums with a raw type like `String` or `Int`
+conform to `RawRepresentable` and we added a default [implementation] for these
+cases. These types of enums can be assigned the raw value that represents each
+case or Swift will automatically assign a raw value for you. In this instance,
+each case has a raw value of a `String` that is that case name, like `case
+User = "User"`. With an `Int`, Swift would automatically assign the top case to
+`0` and increment by one for each case afterward.
```
gfontenot: this paragraph is a bit hard to read.

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,69 @@
...
+    // First, make sure JSON is a number.
+    case let .Number(n):
+
+      // Next, match the number to the enum case.
+      switch n {
+
+      // When a case matches, use pure to wrap the enum in Decoded.
+      case 13.1: return pure(.HalfMarathon) // 13.1 miles is a half marathon distance.
+      case 26.2: return pure(.Marathon) // 26.2 miles is a marathon distance.
+      case _ where n > 26.2: return pure(.UltraMarathon) // Any distance over a marathon is considered an ultra marathon.
```
gfontenot: do we need the comments here?

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,69 @@
...
+
+We can write `Decodable` for `FootRace` like so:
+
+```swift
+extension FootRace: Decodable {
+  static func decode(j: JSON) -> Decoded<FootRace> {
+    switch j {
+
+    // First, make sure JSON is a number.
+    case let .Number(n):
```
gfontenot: how about naming this `distance` instead of `n`. I think that will make things clearer below without the need for comments.

Documentation/Decode-Enums.md
tonyd256: haha not at all. I originally had it without but someone commented that they would rather have the comments.

Documentation/Decode-Enums.md
tonyd256: I think it is suppose to be `is` because of `Decoding` being the verb. I think `Decoding ... is a similar process.` where the object is singular as opposed to `Structs and Classes are similar things` where the object is plural ... however, my english isn't world renown so I will default to @gabebw s thoughts.

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+"THAT'S IT?! How?", you ask. Enums with a raw type like `String` or `Int`
+conform to `RawRepresentable`. Enums that are `RawRepresentable` are given a
+default value by the Swift compiler. With `Int`, the first case is `0` and each
+case is one more than the previous case. With `String`, the default raw value
+is the case name. So for `Role`, the raw value that represents the case `User`
+is `"User"`. `RawRepresentable` enums also get a default initialize
+`init(rawValue: )`. We added a default [implementation] to Argo for `Int` and
+`String` enums so decoding can be this simple.
```
tonyd256: @gfontenot is this better?

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,70 @@
...
+  static func decode(j: JSON) -> Decoded<FootRace> {
+    switch j {
+
+    // First, make sure JSON is a number.
+    case let .Number(distance):
+
+      // Next, match the number to the enum case.
+      switch distance {
+
+      // When a case matches, use pure to wrap the enum in Decoded.
```
tonyd256: @gfontenot do you think a description after the code is better? See this commit: 667970f844fc77158db3b1dd1dd0e4e35e332c8e

Or having neither is better?

Documentation/Decode-Enums.md
dazmuda: You're saying decoding s+c is similar to decoding root keys, so keep it as is.

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,70 @@
+# Decoding Enums
+
+Decoding Structs and Classes is a similar process, but Enums can mix things up.
+If your Enum inherits from a Raw Type `String` or `Int`, then making it conform
```
gfontenot: Parens around "`String` or `Int`" maybe?

Documentation/Decode-Enums.md
```diff
@@ -0,0 +1,70 @@
...
+extension Role: Decodable { }
+```
+
+"THAT'S IT?! How?", you ask. Enums with a raw type like `String` or `Int`
+conform to `RawRepresentable`. Enums that are `RawRepresentable` are given a
+default value by the Swift compiler. With `Int`, the first case is `0` and each
+case is one more than the previous case. With `String`, the default raw value
+is the case name. So for `Role`, the raw value that represents the case `User`
+is `"User"`. `RawRepresentable` enums also get a default initialize
+`init(rawValue: )`. We added a default [implementation] to Argo for `Int` and
```
gfontenot: Maybe make this link "default implementation" instead of just "implementation:?

Documentation/Decode-Enums.md
gfontenot: Yep

Documentation/Decode-Enums.md
gfontenot: I don't think we need the comments for the individual cases. I think the rest are probably fine.

gfontenot:
This looks good to me.

Issue #211, opened 2015-09-04
Decoded should support custom error states


jshier:
In all of the custom decoding I've had to do with the backend I'm current developing against, I've found myself rather limited by `.TypeMismatch` and `.MissingKey` being the only `Decoded` error states. I've started just returning `.TypeMismatch` with a custom string describing the issue, since `.TypeMismatch` comes closest to representing a custom decoding error. For example, if a `double` is being returned as a string and I want to convert it to be an actual `double`, I do this: 
```swift
let toDouble: String -> Decoded<Double> = {
    if let double = Double($0) {
        return pure(double)
    }
    else {
        return .TypeMismatch("Could not convert String \($0) to Double.")
    }
}
```
This gives me a little extra detail without having to examine the actual returned JSON. It'd be great if this was a supported state, like `.General` or `.Custom`. Better names are always good. :smile:  

I mentioned my concern in PR #210.

gfontenot:
So this seems easy/reasonable enough. I don't have any real problem with it. My only question is whether it should be a "fatal" error or not?

Basically, we need to consider whether a failed decoding with a custom type should be considered a failure for an optional property. Right now, this is what we do:

```swift
public extension Decoded {
  static func optional<T>(x: Decoded<T>) -> Decoded<T?> {
    switch x {
    case let .Success(value): return .Success(.Some(value))
    case .Failure(.MissingKey): return .Success(.None)
    case let .Failure(.TypeMismatch(x)): return .Failure(.TypeMismatch(x))
    }
  }
```

Basically, we say that a missing key is actually fine, and that a type mismatch is a real failure. My gut says that custom failures should work like type mismatch failures, and so should be treated as "fatal" for optional values.

jshier:
I agree, which is why I used `.TypeMismatch` as my equivalent for a custom error, as it seemed closest in intent, even if it didn't really describe the issues I could see in custom transformers.

Customization of which errors are fatal is yet another, small, issue I have with Argo. It would be nice to, say, decode an array and not have the decode fail because a single element failed to decode. But I'll raise that issue if I ever run into that requirement while using Argo.

gfontenot:
>  It would be nice to, say, decode an array and not have the decode fail because a single element failed to decode.

Different issue, but fwiw that can be achieved on a case by case basis by using a different decoder.

[...]

PR #237, opened 2015-09-21
Add a Custom error case for DecodeError

tonyd256:
This is to help users with error messaging that doesn't really fit into a type mismatch or missing key. See #211.

### DIFF
--- a/Argo/Types/DecodeError.swift
+++ b/Argo/Types/DecodeError.swift
@@ -1,6 +1,7 @@
 public enum DecodeError: ErrorType {
   case TypeMismatch(expected: String, actual: String)
   case MissingKey(String)
+  case Custom(String)
 }
 
 extension DecodeError: CustomStringConvertible {
@@ -8,6 +9,7 @@ extension DecodeError: CustomStringConvertible {
     switch self {
     case let .TypeMismatch(expected, actual): return "TypeMismatch(Expected \(expected), got \(actual))"
     case let .MissingKey(s): return "MissingKey(\(s))"
+    case let .Custom(s): return "Custom(\(s))"
     }
   }
 }
--- a/Argo/Types/Decoded.swift
+++ b/Argo/Types/Decoded.swift
@@ -16,6 +16,7 @@ public extension Decoded {
     case let .Success(value): return .Success(.Some(value))
     case .Failure(.MissingKey): return .Success(.None)
     case let .Failure(.TypeMismatch(x)): return .Failure(.TypeMismatch(x))
+    case let .Failure(.Custom(x)): return .Failure(.Custom(x))
     }
   }
 
@@ -38,6 +39,10 @@ public extension Decoded {
 
   static func missingKey<T>(name: String) -> Decoded<T> {
     return .Failure(.MissingKey(name))
+  }
+
+  static func customError<T>(message: String) -> Decoded<T> {
+    return .Failure(.Custom(message))
   }
 }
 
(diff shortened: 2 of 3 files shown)


paulyoung:
Looks good to me.

Not that I can think of a use for it right now, but I think this would also make `materialize` possible if someone really wanted it (I explained in #225 how I couldn't make the catch exhaustive).

jshier:
Seems like there should be a `Custom` equivalent of `.typeMismatch`, no? `.custom` or `.customError` should work.

tonyd256:
@jshier I passed on the helper function because it's just a custom `String` that the user generates. So you can use `.Custom("")` to make it. If you think a `.custom(String)` helper function would be helpful then I could add it. I could also be convinced to have it for consistency and completeness :)

jshier:
I like `.typeMismatch` because it's slightly easier and more elegant to call rather than `.Failure(.TypeMismatch(expect:"", actual:""))`. While `.Failure(.Custom("String"))` isn't quite as bad, I do think `.custom("String")` is a bit better. And yes, it would feel a bit odd to have convenience functions for the other error states but not `Custom`.

tonyd256:
Ah right, forgot about `.Failure` needing to wrap that ... ok adding ... brb

tonyd256:
Figured `.customError` was a bit more explicit of what it was doing than just a function named `custom` on the `Decoded` type.

jshier:
Thanks, that looks great.

gfontenot:
:+1: from me

edwardaux:
This looks great. I was just about to raise an issue asking if anyone else had ever needed this :)

(although I was going to suggest using `.InvalidData`, but `.Custom` gives ultimate flexibility)

edwardaux:
@tonyd256 am I going mad, or did this merge get lost somehow? The timeline above says that 038f72a was merged into `master`, but https://github.com/thoughtbot/Argo/blob/master/Argo/Types/DecodeError.swift definitely doesn't contain it.

jshier:
Commit history for master doesn't show it either.

edwardaux:
I have a project that checked it out as a submodule (see screenshot below), but unfortunately my `Cartfile` specifically referenced the commit hash, not the branch, so I can't tell if it ever made it to `master`.

screen shot 2015-09-30 at 9 02 03 am

If you contrast that to the current history, it is nowhere to be found :scream: 

screen shot 2015-09-30 at 9 05 12 am

tonyd256:
ugh ... crap ... @jnutting or @gfontenot did either of you happen to remove this commit when merging in your branches today?

jnutting:
I... don't know!? Surely not on purpose. I suppose it's possible that I did it on accident, but how would I even know? I don't know. Maybe you merged this roughly the same time that I rebased and merged my documentation changes? Maybe I did a force-push from exactly the wrong feature branch at exactly the wrong time?

gfontenot:
Weird. I definitely didn't force-push master, no idea where this would have gone. I'm going to cherry-pick the commit back into master now.

gfontenot:
Actually, I don't have that commit locally. No idea what's going on here. If anyone _does_ have that commit locally and they want to cherry-pick it into master then open a PR I'd be eternally grateful.

tonyd256:
I have the commit locally on my personal Mac. I'll have to do it tonight when I get home.

edwardaux:
@tonyd256 just in case they've been lost on your local machine, I've created a gist that contains two patches - one for each lost commit.

https://gist.github.com/edwardaux/93e3108ae0dc43c1c3e7

tonyd256:
@paulyoung @edwardaux Fixed now!

paulyoung:
@tonyd256 did you need to force push to `master` to fix this?

tonyd256:
yes ... rebases on open PRs will be harry ... but once they're done we'll be good to go.

edwardaux:
Thanks @tonyd256. Works like a champ!

Issue #141, opened 2015-06-05
Make decoding from a root key easier


tonyd256:
Currently to decode something in a root key like this:
```
{
  "user":
  {
    "id": 5,
    "name": "Bob"
  }
}
```
You need to do extra work to pull out the `User` object from the root key like so:
```swift
let json: AnyObject? = NSJSONSerialization.JSONObjectWithData ...
let user: User? = json >>- decode >>- { ($0 <| "user").value }
let decodedUser: Decoded<User> = json >>- decode >>- { $0 <| "user" } 
```

The purpose of the `decode` function was to make it easy to decode the JSON into the model without having to worry about the operators or using different code to go between a `Decoded` or `Optional` return type.

However, this breaks down with root keys. I think this feels weird and ruins the purpose of the `decode` function. How can we fix this? I have 3 propositions.

Overload all the `decode` operators to take a root key making the API look like so:
```swift
let user: User? = json >>- decode(rootKey: "user") // with root key
let user: User? = json >>- decode // without root key
```
I think this is super easy for the user to type, but adds a lot of code to the library.

Add the rootKey as a parameter on all decode functions:
```swift
let user: User? = json >>- decode(rootKey: "user") // with root key
let user: User? = json >>- decode(rootKey: .None) // without root key
```
I think this is not too much more work for the user but is a little weird. Could also make the `rootKey` label not required but then I think that might look confusing.

Add a new function for pulling out rootKeys:
```swift
let user: User? = json >>- rootKey("user") >>- decode // with root key
let user: User? = json >>- decode // without root key
```
This is also not so much work for the user and things don't change if you don't have root keys (like the first option). It also adds minimal code to the repo. I imagine it would look like this:
```swift
public func rootKey(key: String)(object: AnyObject) -> AnyObject? {
  return (object as? [String: AnyObject])?[key]
}
```

I'm also open to other suggestions.

alskipp:
Do you think the following could be a possible solution?
```swift
func decode<T : Decodable where T == T>(rootKey: String? = .None)(object: AnyObject) -> T?
```
```swift
let user: User? = json >>- decode(rootKey: "user") // with root key
let user: User? = json >>- decode // without root key
```

tonyd256:
@alskipp Let me know what you find ... @gfontenot suggested this too but I found it not possible.

alskipp:
I'm not having any luck getting that idea to work. Doesn't seem possible currently, maybe it will be next week with Swift 2.0?

[...]

PR #238, opened 2015-09-21
Add `decode` function with root keys

tonyd256:
The current `decode` function can't decode models from JSON where the
JSON for that model is embedded within a root key. This is a very common
thing so this is an attempt to make it easier to decode those objects.

This solves #141 and might be a better solution than 8483513b9340025729bbf396c85ea4ac532c160d

### DIFF
--- a/Argo/Functions/decode.swift
+++ b/Argo/Functions/decode.swift
@@ -1,3 +1,5 @@
+// MARK: decode root objects and arrays
+
 public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) -> T? {
   return decode(object).value
 }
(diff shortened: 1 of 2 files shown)


Argo/Functions/decode.swift
```diff
@@ -13,3 +15,21 @@ public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) ->
 public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) -> Decoded<[T]> {
   return Array<T>.decode(JSON.parse(object))
 }
+
+// MARK: decode with a root key
+
+public func decode<T: Decodable where T == T.DecodedType>(rootKey: String, _ object: AnyObject) -> T? {
```
tonyd256: I wanted to make the top 4 `decode` functions like this: 
`public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject, rootKey: String? = .None) -> Decoded<[T]>`
but this breaks partial application. So I decided to make overloads that take non-optional strings.

gfontenot: I feel like this is backwards. `decode(object, rootKey: "foo")` reads much nicer IMO than `decode("foo", object)`

tonyd256: I agree, I initially had it that way but I switched it up for partial application purposes. Check out the tests that changed in this PR. It's nice to be able to partially apply the root key then pass in the `AnyObject`.

tonyd256: I could possibly change the name so it reads better: `decodeWithRootKey` instead.

jnutting: I like the idea of renaming it as you suggest.

tonyd256: @gfontenot thoughts on those suggestions?

gfontenot: Yeah, I guess I'm ok with `decodeWithRootKey`.

PR #242, opened 2015-09-25
Attempt to improve the overall structure and flow of this document.

jnutting:
Reworked and rewrote some things.
Added some explanatory text here and there.
Added subsection headers to partition things a bit.
Updated code snippets to Swift 2 and to latest Argo usage.

### DIFF
--- a/Documentation/Basic-Usage.md
+++ b/Documentation/Basic-Usage.md
@@ -1,9 +1,9 @@
 ## Usage
 
 Argo uses Swift's type system along with concepts from functional programming to
-let you smoothly transform JSON data into Swift model objects or structs. Argo's
-approach does this with a minimum of syntax, while at the same time improving
-type safety and data integrity compared to other approaches.
+let you smoothly transform JSON data into Swift model types. Argo does this with
+a minimum of syntax, while at the same time improving type safety and data
+integrity compared to other approaches.
 
 You may need to learn a few things in order to learn Argo effectively, but once
 you do so, you'll have a powerful new tool to hang on your belt!
@@ -11,10 +11,10 @@ you do so, you'll have a powerful new tool to hang on your belt!
 ### Decoding basics
 
 Argo's whole purpose is to let you easily pick apart structured data (normally
-in the form of a dictionary created from JSON data) and create Swift objects or
-structs based on the decoded content. Typically, you'll want to do this with
-JSON data received from a server or elsewhere. The first thing you need to do is
-convert the JSON data from `NSData` to an `AnyObject` using `Foundation`'s
+in the form of a dictionary created from JSON data) and create Swift objects
+based on the decoded content. Typically, you'll want to do this with JSON data
+received from a server or elsewhere. The first thing you need to do is convert
+the JSON data from `NSData` to an `AnyObject` using `Foundation`'s
 `NSJSONSerialization` API.  Once you have the `AnyObject`, you can call Argo's
 global `decode` function to get back the decoded model.
 
@@ -38,9 +38,8 @@ return type as seen in the code block above.
 
 ### Implementing `Decodable`
 
-In order for this to work with your own model classes or structs, you need to make
-sure that models that you wish to decode from JSON conform to the `Decodable`
-protocol:
+In order for this to work with your own model types, you need to make sure that
+models that you wish to decode from JSON conform to the `Decodable` protocol:
 
 ```swift
 public protocol Decodable {
@@ -51,7 +50,7 @@ public protocol Decodable {
 
 In your model, you need to implement the `decode` function to perform whatever
 transformations are needed in order to create your model from the given JSON
-structure.  To illustrate this, we will decode a simple model object called
+structure.  To illustrate this, we will decode a simple model type called
 `User`. Start by creating this `User` model:
 
 ```swift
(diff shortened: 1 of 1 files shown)


Documentation/Basic-Usage.md
```diff
@@ -1,44 +1,58 @@
...
+`Decoded` type and use it to debug or report decoding failures.  When you decode
+an `AnyObject` into a model using the global `decode` function, you can specify
+whether you want an `Optional` model or a `Decoded` model by specifying the
+return type as seen in the code block above.
+
+### Implementing `Decodable`
 
-Next, you need to make sure that models that you wish to decode from JSON
-conform to the `Decodable` protocol:
+In order for this to work with your own model classes or structs, you need to make
```
gfontenot: I think we should remove `classes or structs` here and go back to `models`. Or possibly say `model types`. Argo works with `enum`s too, I don't want us to give the impression that it doesn't.

Documentation/Basic-Usage.md
```diff
@@ -47,57 +61,88 @@ struct User {
...
+
+### Safely pulling values from JSON
 
-In order to help with the decoding process, Argo introduces two new operators
-for parsing a value out of the JSON:
+In the example above, we showed some non-existent parameters (`paramOne`, etc), but
+one of Argo's main features is the ability to help you grab the real parameters
+from the JSON structure in a way that is type-safe and concise. You don't need
+to manually check to make sure that a value is non-nil, or that it's of the
+right type. Argo leverages Swift's expressive type system to do that leaving
```
gfontenot: s/leaving/heavy ?

Documentation/Basic-Usage.md
```diff
@@ -47,57 +61,88 @@ struct User {
...
+operators for parsing a value out of the JSON:
 
 - `<|` will attempt to parse a single value from the JSON
 - `<||` will attempt to parse an array of values from the JSON
 
-The usage of these operators is the same regardless:
+The usage of these operators is simple:
+
+- `json <| "name"` is analogous to `json["name"]`
+- `json <|| "posts"` is analogous to `json["posts"]`
```
gfontenot: Should this make a note that `posts` in this case is expected to be a JSON array?

Documentation/Basic-Usage.md
```diff
@@ -47,57 +61,88 @@ struct User {
...
-value it finds is the wrong type, the function will return a
-`Decoded.TypeMismatch(message: String)` error.
+- `json <| ["location", "city"]` is analogous to `json["location"]["city"]`
+
+Each of these operators will attempt to extract the specified value from the
+JSON structure and if a value is found, the operator will also attempt to cast
+the value to the expected type. If it can't find a value, the function will
+return a `Decoded.MissingKey(message: String)` failure state. If the value it
+finds is of the wrong type, the function will return a
+`Decoded.TypeMismatch(message: String)` failure state.
```
gfontenot: These are out of date. The `Decoded` type only has `Success` and `Failure` cases now. The `Failure` case can be `DecodeError.MissingKey(key: String)` or `DecodeError.TypeMismatch(expected: String, actual: String)`.

gfontenot: And there's a PR to add a custom error case to that list.

gfontenot: Maybe we should have an `Understanding failures` document?

Documentation/Basic-Usage.md
```diff
@@ -112,13 +157,14 @@ extension User: Decodable {
...
-For comparison, this same function without Argo would look like so:
+For comparison, an implementation of a similar function without Argo could look
+like this:
 
 ```swift
 extension User {
   static func decode(j: NSDictionary) -> User? {
-    if let id = j["id"] as Int,
-       let name = j["name"] as String
+    if let id = j["id"] as! Int?,
```
gfontenot: I think this should be `if let id = j["id"] as? Int`. Ditto below.

Documentation/Basic-Usage.md
```diff
@@ -128,7 +174,13 @@ extension User {
 }
 ```
 
-You could see how this would get much worse with a more complex model.
+Not only is that code much more verbose than the equivalent code using Argo,
+it's also not as safe. It does check to make sure that `id` and `name` are
+non-nil, but it's not checking to see that their types are correct. If that code
+encountered JSON where the `"id"` key returned, say a `String` instead of an
+`Int`, this would lead to a crash.
```
gfontenot: This paragraph feels like a bit of a straw man argument. I think that better arguments would be: lack of failure information, verbosity, and duplicate type information (if the type of `id` changes to a String, you have to change that in multiple places).

gfontenot: Also, maybe not for this doc, but this _only_ works for std lib types. You can't just cast to `MyCoolRelationshipType` and have it work.

gfontenot:
Some comments, but this is rad. Pinging @thoughtbot/copy-editing as well.

Documentation/Basic-Usage.md
```diff
@@ -1,44 +1,58 @@
 ## Usage
 
-The first thing you need to do when you receive JSON data is convert it from
-`NSData` to an `AnyObject` using `Foundation`'s `NSJSONSerialization` API.
-Once you have the `AnyObject`, you can call the global `decode` function to get
-back the decoded model.
+Argo uses Swift's type system along with concepts from functional programming to
+let you smoothly transform JSON data into Swift model objects or structs. Argo's
+approach does this with a minimum of syntax, while at the same time improving
```
gabebw: Instead of "Argo's approach", maybe just "Argo"?

mcmire:
I was able to easily follow all of that as a non-Swift developer, so nice job. :star:

Documentation/Basic-Usage.md
```diff
@@ -47,57 +61,88 @@ struct User {
...
+the course of the decoding process. This basically means that we can build up
+the `init` function call bit by bit, adding one parameter at a time, if and only
+if Argo can successfully decode them. If any of the parameters don't meet our
+expectations, Argo will skip the `init` call and return a special failure state. 
 
 [`Curry`]: https://github.com/thoughtbot/Curry
 
-- [Apple's documentation of curried functions](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/Declarations.html#//apple_ref/doc/uid/TP40014097-CH34-XID_615)
-- [Introduction to Function Currying in Swift](http://robots.thoughtbot.com/introduction-to-function-currying-in-swift)
+> If you'd like to learn more about currying, we recommend the following articles:
```
gabebw: This is a nice formatting change.

Documentation/Basic-Usage.md
gabebw: Do you mean JSON object, not array?

Documentation/Basic-Usage.md
```diff
@@ -47,57 +61,88 @@ struct User {
...
+to manually check to make sure that a value is non-nil, or that it's of the
+right type. Argo leverages Swift's expressive type system to do that leaving
+lifting for you. To help with the decoding process, Argo introduces two new
+operators for parsing a value out of the JSON:
 
 - `<|` will attempt to parse a single value from the JSON
 - `<||` will attempt to parse an array of values from the JSON
 
-The usage of these operators is the same regardless:
+The usage of these operators is simple:
```
gabebw: How about: "These are infix operators for familiar operations"?

Documentation/Basic-Usage.md
```diff
@@ -47,57 +61,88 @@ struct User {
...
+
+- `json <| "name"` is analogous to `json["name"]`
+- `json <|| "posts"` is analogous to `json["posts"]`
 
-- `json <| "key"` is analogous to `json["key"]`
-- `json <| ["key", "nested"]` is analogous to `json["key"]["nested"]`
+As a bonus, if your JSON contains nested data whose intermediate elements are
```
gabebw: I think "intermediate" is not the right word, but I can't think of the right one. Maybe just remove it? It already says "nested data".

gabebw:
These changes are really readable. Great job, Jack.

Documentation/Basic-Usage.md
gfontenot: Well, a JSON array of JSON objects

gfontenot:
```json
{
  "posts": [
    { "foo": "bar"}
  ]
}
```

Documentation/Basic-Usage.md
jnutting: I don't agree that mentioning type-safety is a straw-man argument (that's a very real thing that would likely even lead to a crash in objc code which might otherwise at least cope with a nil value), but the other points you raise are also important. I'll try to find some way to mention all of those.

jnutting: Or, wait, of course `as?` is in fact doing type-checking, so nvm.

PR #243, opened 2015-09-25
Extend CollectionType intead of extending Array

gfontenot:
Extending Array was great, but it'd be nicer if we could extend _any_
collection type that contains Decodable objects. This `decode` function
will still always return an Array, for the same reason that the stdlib's
`map`, `flatMap`, etc return Arrays, but if we ever get Real Actual
Higher Kinded Types, we can fix this.

Note that this _almost_ works with Dictionary as well, except that it would
always return an array, instead of returning a dictionary. Bummer.

### DIFF
--- a/Argo/Types/StandardTypes.swift
+++ b/Argo/Types/StandardTypes.swift
@@ -60,10 +60,10 @@ public extension Optional where Wrapped: Decodable, Wrapped == Wrapped.DecodedTy
   }
 }
 
-public extension Array where Element: Decodable, Element == Element.DecodedType {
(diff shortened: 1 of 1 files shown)


paulyoung:
In case this is of any use whatsoever regarding dictionaries: https://github.com/edwardaux/Ogra/pull/1

paulyoung:
Sorry, the diff of that is probably the most useful: https://github.com/edwardaux/Ogra/pull/1/files

tonyd256:
horse_love_it

PR #244, opened 2015-09-25
Add tests for decoding dictionaries

gfontenot:
Turns out, we weren't testing the decoding of dictionaries anywhere.
That seems wrong, so this adds those tests and (thankfully) proves that
our current implementation works as expected.

Also, we ran into a swift bug that I actually hit the other day as well
that I need to file a radar on. The title of which will probably end up
being:

> flatMapping custom monadic type with static function defined in a
> protocol extension results in compiler crash

That's why we're duplicating the meat of `flatMap` for the purpose of
the test.

### DIFF
--- a/ArgoTests/Tests/PListDecodingTests.swift
+++ b/ArgoTests/Tests/PListDecodingTests.swift
@@ -20,5 +20,6 @@ class PListDecodingTests: XCTestCase {
     XCTAssert(model?.eStringArrayOpt?.count == 0)
     XCTAssert(model?.userOpt != nil)
     XCTAssert(model?.userOpt?.id == 6)
+    XCTAssert(model!.dict == ["foo": "bar"])
   }
 }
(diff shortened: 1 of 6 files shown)


paulyoung:
Is there a reason for using `!` when the other assertions use `?` ?

gfontenot:
That's what we had to do to get it to compile :confused:

I don't use bare `XCTest` enough to know if there's a better way to solve that.

ArgoTests/Tests/PListDecodingTests.swift
```diff
@@ -20,5 +20,6 @@ class PListDecodingTests: XCTestCase {
     XCTAssert(model?.eStringArrayOpt?.count == 0)
     XCTAssert(model?.userOpt != nil)
     XCTAssert(model?.userOpt?.id == 6)
+    XCTAssert(model!.dict == ["foo": "bar"])
```
tonyd256: these probably need to be `!` instead of `?` because native Swift comes with `==` implementations of `Optional<Type> == Type` for all it's native types. They probably didn't have one for `Dictionary` ... might be able to just wrap the Dictionary in `.Some` or `pure` if Runes is imported above.

tonyd256:
One comment but otherwise good to merge

PR #246, opened 2015-09-27
Add tvOS target

nickygerritsen:
Took me some while to figure out that I need the xcworkspace for the tests to succeed.

Anyway, this should be a tvOS version with the scheme unshared.


nickygerritsen:
Oh yeah the podfile has not been updated yet, as I think that they do not have official tvOS support yet.

(btw this also holds for my Curry tvOS PR)

Cartfile.private
```diff
@@ -1,2 +1,2 @@
-github "thoughtbot/Curry"
+github "thoughtbot/Curry" "tvos"
 github "thoughtbot/Runes" >= 3.0.0
```
bruceflowers: @nickygerritsen  you can now point to this branch of Runes below, as https://github.com/thoughtbot/Runes/pull/51 has been merged and there's a new tvos branch that shares the scheme:

github "thoughtbot/Runes" "tvos"

nickygerritsen: But the runes code is directly added to the Argo project so we don't actually use any Runes targets

gfontenot: Correct. We don't need this to point at the `tvos` branch of Runes since we aren't building the targets.

nickygerritsen:
Now that XCode 7.1 GM is live, I guess we can share the scheme again, and fix the cocoapod (as they seem to support tvOS now too), right?

I'm not sure if I can do this anytime soon as I'll be leaving on holiday next week and have a lot to do, so perhaps someone else can do these last steps :)?

tonyd256:
Merged in here: cb88f764443571f3ce71bc3ed4ae7ba06943b714

PR #247, opened 2015-09-27
Extend CollectionType instead of Dictionary

gilesvangruisen:
> Similar to #243, in which we extended CollectionType instead of
Array to parse any collection of Decodable obects. That, however,
always returns an Array, so here we extend CollectionType again
(for cases where Self is DictionaryLiteralConvertible) and instead
return a Dictionary of Decoded objects.

Ideally, as mentioned in #243, we'd be able to wrap this into a single extension on `CollectionType` that can handle both `Array` and `Dictionary`. Until then, this replaces the `Dictionary` extension and lets you parse any `DictionaryLiteralConvertible` collection of `Decodable` objects.

This also passes the dictionary decoding tests in #244

Thanks @paulyoung for sharing the [Ogra diff](https://github.com/edwardaux/Ogra/pull/1/files) that inspired this change

### DIFF
--- a/Argo/Types/StandardTypes.swift
+++ b/Argo/Types/StandardTypes.swift
@@ -69,10 +69,10 @@ public extension Array where Element: Decodable, Element == Element.DecodedType 
   }
 }
 
-public extension CollectionType where Self: DictionaryLiteralConvertible, Self.Value: Decodable, Self.Value == Self.Value.DecodedType {
-  static func decode(j: JSON) -> Decoded<[String: Self.Value]> {
+public extension DictionaryLiteralConvertible where Value: Decodable, Value == Value.DecodedType {
+  static func decode(j: JSON) -> Decoded<[String: Value]> {
     switch j {
-    case let .Object(o): return sequence(Self.Value.decode <^> o)
+    case let .Object(o): return sequence(Value.decode <^> o)
     default: return .typeMismatch("Object", actual: j)
     }
   }
--- a/Argo/Types/StandardTypes.swift
+++ b/Argo/Types/StandardTypes.swift
@@ -69,10 +69,10 @@ public extension Array where Element: Decodable, Element == Element.DecodedType 
   }
 }
 
-public extension Dictionary where Value: Decodable, Value == Value.DecodedType {
-  static func decode(j: JSON) -> Decoded<[String: Value]> {
+public extension CollectionType where Self: DictionaryLiteralConvertible, Self.Value: Decodable, Self.Value == Self.Value.DecodedType {
+  static func decode(j: JSON) -> Decoded<[String: Self.Value]> {
     switch j {
-    case let .Object(o): return sequence(Value.decode <^> o)
+    case let .Object(o): return sequence(Self.Value.decode <^> o)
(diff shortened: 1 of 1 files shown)


Argo/Types/StandardTypes.swift
```diff
@@ -69,10 +69,10 @@ public extension Array where Element: Decodable, Element == Element.DecodedType
   }
 }
 
-public extension Dictionary where Value: Decodable, Value == Value.DecodedType {
-  static func decode(j: JSON) -> Decoded<[String: Value]> {
+public extension CollectionType where Self: DictionaryLiteralConvertible, Self.Value: Decodable, Self.Value == Self.Value.DecodedType {
```
gfontenot: is `Self.` needed? Or can it be made implicit?

gilesvangruisen: I could be wrong but I think we can only take advantage of the implicit `Self.` when adding constraints to types that directly belong to the type or protocol that is being extended (in this case `CollectionType`), and not when adding constraints to types that belong to another constraint on the extended type. Does that make sense?

Because we're adding `DictionaryLiteralConvertible` as a separate constraint on `Self`, the `Self.` is necessary to constrain types belonging to `DictionaryLiteralConvertible.

gilesvangruisen:
Another way to do this would be to extend `DictionaryLiteralConvertible` directly:
```Swift
public extension DictionaryLiteralConvertible where Value: Decodable, Value == Value.DecodedType
```

This *should* be functionally equivalent since any `DictionaryLiteral` inherently conforms to `CollectionType`.

The dictionary tests pass just fine but it might be worthwhile to write a spec that tests `CollectionType` beyond just `Array` and `Dictionary`

Argo/Types/StandardTypes.swift
paulyoung: I found that I needed to use `Self.` when making the changes to Ogra.

tonyd256:
@gilesvangruisen I know you have a code snippet above, but would you mind making another commit on top of this that extends `DictionaryLiteralConvertable` instead? I'm curious what it looks like in comparison.

gilesvangruisen:
@tonyd256 Sure thing. It actually cleans up the body too since we can omit `Self.`

paulyoung:
Looks good to me :+1:

paulyoung:
Actually, maybe a dumb question but do we now need to specify type constraints for `Key`?

I assumed that it was `DictionaryLiteralConvertible` that required keys to be `Hashable` but apparently not: http://swiftdoc.org/v2.0/protocol/DictionaryLiteralConvertible/

gfontenot:
> do we now need to specify type constraints for `Key`

I don't think so, because we're really only dealing with `String` keys, but maybe I'm wrong? It's possible that we should restrict that explicitly?

tonyd256:
I think this is fine for now. It's passing our tests and working. If we fine a case where it breaks then we should revisit.

tonyd256:
Thanks for this! Merged in here: 3a3a7939269d3c2a63e6e12cfe72f4ca50a4f379

gilesvangruisen:
Awesome, thanks!

PR #250, opened 2015-09-30
Add catDecoded and ?? to Decoded

gfontenot:
- `??` works the same way that you'd expect. If the `lhs` argument is a
   `.Failure` case, it returns the default value on the right. If it's a
  `.Success`, it returns the unwrapped value.
 - `catDecoded` takes an array of `Decoded<T>` objects, and returns the
   unwrapped `.Success` values, discarding any `.Failure` values. The
   same function is implemented for dictionaries.

This isn't needed in our default API, but enough people are looking for
this kind of functionality that it makes sense for us to provide it
officially.

### DIFF
--- a/Argo/Operators/DecodedOperators.swift
+++ b/Argo/Operators/DecodedOperators.swift
@@ -83,9 +83,9 @@ public func <|><A>(lhs: Decoded<A>, rhs: Decoded<A>) -> Decoded<A> {
   - returns: A value of type `A`
 */
 
-public func ?? <A>(lhs: Decoded<A>, rhs: A) -> A {
+public func ?? <A>(lhs: Decoded<A>, @autoclosure rhs: () -> A) -> A {
   switch lhs {
   case let .Success(x): return x
-  case .Failure: return rhs
+  case .Failure: return rhs()
   }
 }
--- a/Argo/Operators/DecodedOperators.swift
+++ b/Argo/Operators/DecodedOperators.swift
@@ -68,3 +68,24 @@ public func <|><A>(lhs: Decoded<A>, rhs: Decoded<A>) -> Decoded<A> {
   }
   return rhs
 }
+
+// MARK: Failure coalescing operator
+
+/**
+  return the unwrapped value of the `Decoded` value on the left if it is `.Success`, otherwise return the default on the right
+
+  - If the left hand side is successful, this will return the unwrapped value from the left hand side argument
+  - If the left hand side is unsuccesful, this will return the default value on the right hand side
+
+  - parameter lhs: A value of type `Decoded<A>`
+  - parameter rhs: A value of type `A`
+
+  - returns: A value of type `A`
+*/
+
+public func ?? <A>(lhs: Decoded<A>, rhs: A) -> A {
+  switch lhs {
+  case let .Success(x): return x
+  case .Failure: return rhs
+  }
+}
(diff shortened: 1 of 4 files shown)


Argo/Types/Decoded.swift
```diff
@@ -85,3 +85,10 @@ public extension Decoded {
     }
   }
 }
+
+public func ?? <T>(lhs: Decoded<T>, rhs: T) -> T {
```
tonyd256: we should stick this where we stuck Alternative `<|>` ... I think in `DecodedOperators.swift`

gfontenot: Yeah, wasn't sure. I feel like we need to reorganize all of the operators into different files.

gfontenot: I'll make that change though.

tonyd256: ya I agree ... I don't like those file names but I didn't have a better idea at the time ... I'm open to suggestions

gfontenot:
@tonyd256 updated

Argo/Operators/DecodedOperators.swift
```diff
@@ -68,3 +68,24 @@ public func <|><A>(lhs: Decoded<A>, rhs: Decoded<A>) -> Decoded<A> {
...
+  - If the left hand side is successful, this will return the unwrapped value from the left hand side argument
+  - If the left hand side is unsuccesful, this will return the default value on the right hand side
+
+  - parameter lhs: A value of type `Decoded<A>`
+  - parameter rhs: A value of type `A`
+
+  - returns: A value of type `A`
+*/
+
+public func ?? <A>(lhs: Decoded<A>, rhs: A) -> A {
```
ikesyo: `rhs` should be marked with `@autoclosure`, I think.

tonyd256: In Swift 2, these are the definitions for the Optional `??`:

```swift
@warn_unused_result
@rethrows public func ??<T>(optional: T?, @autoclosure defaultValue: () throws -> T?) rethrows -> T?

@warn_unused_result
@rethrows public func ??<T>(optional: T?, @autoclosure defaultValue: () throws -> T) rethrows -> T
```
The first one being the same as what we consider Alternative, `<|>`. Since `Decoded` is our form of error handling, I don't think we need any of the `throws` or `rethrows` stuff. We should probably add `@autoclosure` to alternative as well. AND AND `@noescape` to all the other monadic operators for `Decoded`. Another PR though.

tonyd256:
@gfontenot I'm thinking about cutting a release tomorrow. Would you be able to get this in?

Argo/Operators/DecodedOperators.swift
gfontenot: Good call @ikesyo, fixed.

I'm not going to worry about the rethrows stuff right now. We should look into that in a different PR. I'll create an issue to track it.

gfontenot:
@tonyd256 Was there anything else that you thought needed to happen here? I can merge if not.

Argo/Operators/DecodedOperators.swift
```diff
@@ -68,3 +68,24 @@ public func <|><A>(lhs: Decoded<A>, rhs: Decoded<A>) -> Decoded<A> {
...
+// MARK: Failure coalescing operator
+
+/**
+  return the unwrapped value of the `Decoded` value on the left if it is `.Success`, otherwise return the default on the right
+
+  - If the left hand side is successful, this will return the unwrapped value from the left hand side argument
+  - If the left hand side is unsuccesful, this will return the default value on the right hand side
+
+  - parameter lhs: A value of type `Decoded<A>`
+  - parameter rhs: A value of type `A`
```
tonyd256: do you think this should change since it now has an `@autoclosure`?

tonyd256:
One comment then :dancer:

Argo/Operators/DecodedOperators.swift
gfontenot: yep

Issue #252, opened 2015-10-02
Be consistent with generic naming


gfontenot:
We're using `A` and `B` in some places, and `T` and `U` in others. This is driving me insane. We should fix that.

PR #258, opened 2015-10-02
Standardize generic types and variables

tonyd256:
It seems that Swift is standardizing on T and U for generic parameters
over Haskell's a and b so let's follow that. Use V if you need a third
and S for sequences. Also, let's standardize on math variables x and y.


tonyd256:
fixes #252

gfontenot:
:+1: yes please

Issue #254, opened 2015-10-02
Move `pure` with `apply` and `flatMap`


gfontenot:
`pure` is currently in `Decoded.swift`, which is weird since it's part of `Applicative` and `Monad`. We should move it to be alongside `<*>` and `>>-`.

Issue #253, opened 2015-10-02
Organize operators more logically


gfontenot:
We have `Operators`, `DecodedOperators` and now `OptionalOperators`. This is confusing. We should find a better way to organize these.

tonyd256:
ugh sorry, wrong button

PR #259, opened 2015-10-02
Re-organize operators into separate files

tonyd256:
This moves the monadic operators into their own files and renames the
files with the Argo operators to (ReturnType)Pull.swift.


tonyd256:
An attempt at #253 and #254

gfontenot:
I don't think this actually solves #254, right? Am I missing it somewhere?

I'm good with this. I think it's the most reasonable change. However, I question the use of `Pull` as the name for the file containing the `<|` family of operators. I don't _believe_ we've been using the word `pull` for that operator. I thought we had generally settled on `decode` as the word. I'd also probably flip the words so that it became `DecodeOptional` and `DecodeDecoded` (or whatever we actually settle on for the name of `<|`). Seems like they read better, and they'd sort together alphabetically as well.

tonyd256:
That's fine with me. I actually had it that way initially then changed when I read "Pull" in the comments.

tonyd256:
It solves #254 in that it moves `pure` into the `Applicative` file since it is a function on Applicative in Haskell.

gfontenot:
I don't see `pure` in `Applicative.swift`

tonyd256:
@gfontenot yup sorry. I thought I had done it. It's done now.

gfontenot:
dig it

Maybe taking things too far, but: what do you think about moving the implementations of `map`, `apply`, `flatMap` into `Functor.swift`, `Applicative.swift`, and `Monad.swift`? We already have a non-operator in there, and if those are about type conformance, maybe it makes sense? We could also move them out of an `Operators` directory. Maybe `Argo/Decoded/Monad.swift` and `Argo/Decoded/Decoded.swift`?

gfontenot:
if that's insane, feel free to merge because this is a huge step up as is.

tonyd256:
I actually like that idea! Will do! Then will merge!

tonyd256:
@gfontenot Done. Didn't merge yet. Want to make sure it looks good.

Argo/Types/Decoded/Decoded.swift
```diff
@@ -54,7 +54,7 @@ public extension Decoded {
 }
 
 extension Decoded: CustomStringConvertible {
-  public var description: String {
+   public var description: String {
```
tonyd256: ugh sorry

gfontenot:
I think this is a great change. Let's merge this shit.

Issue #248, opened 2015-09-29
throws clause in create() causes "ambiguous reference to member" error


edwardaux:
For some reason, having a `throws` clause in my curried `create` method causes the Swift compiler some problems trying to resolve the `<|` operator. Consider the following code:

	enum Blah {
		case XXX(String, String)
	}
	extension Blah : Decodable {
		static func create(type: String)(parameter: String) throws -> Blah {
			return .XXX(type, parameter)
		}
		
		static func decode(j: JSON) -> Decoded<Blah> {
			return Blah.create
				<^> j <|  "type"
				<*> j <| "parameter"
		}
	}

Removing `throws` makes the error go away (see error screenshot below). However, what is not shown is that I need to do some validation in the `create` function which may mean an error gets thrown. Note that I used to return a `Result<Blah, NSError>` object and everything worked OK, but I'm slowly moving towards using `throws` as my error handling technique. Do I need to go back to using `Result` for this piece?

![screen shot 2015-09-29 at 4 57 11 pm](https://cloud.githubusercontent.com/assets/1574429/10157202/32cc0eac-66cb-11e5-93c3-be36ba7360d2.png)

tonyd256:
this makes sense because adding `throws` in the `create` function signature changes the type. Argo's operators are not setup to handle `throws`. I could use a `try-catch` inside the `create` function and inside the `catch` return a custom `DecodedError` instead.

tonyd256:
Actually scratch that. I would abstract a function that performs the throwing code say `readFile()`. then have that function return a `DecodedError` if the `try` fails:

```swift
func readFile(path: String) -> Decoded<String> {
  do {
    let string = try readFileWithPath(path) // function that could throw
    return pure(string)
  } catch {
    return .custom("Custom Error String")
  }
}
```
Then in your `decode` function you would use it like so: `<*> (j <| "parameter" >>- readFile)`

paulyoung:
Maybe a use case for `materialize` after all?

tonyd256:
hmm oh yes, possibly!

edwardaux:
Thanks @tonyd256, I like the idea of abstracting out the function. It keeps things isolated, and it means I don't have to use a `Result` object just for this isolated instance. Thanks for the feedback and suggestion.

PR #261, opened 2015-10-03
Add materialize.

paulyoung:
https://github.com/thoughtbot/Argo/issues/248#issuecomment-144622463 got me thinking that `materialize` could be useful after all.

cc @edwardaux

### DIFF
--- a/Argo/Types/Decoded/Decoded.swift
+++ b/Argo/Types/Decoded/Decoded.swift
@@ -70,3 +70,11 @@ public extension Decoded {
     }
   }
 }
+
+public func materialize<T>(f: () throws -> T) -> Decoded<T> {
+  do {
+    return .Success(try f())
+  } catch {
+    return .customError("\(error)")
+  }
+}
--- a/ArgoTests/Tests/DecodedTests.swift
+++ b/ArgoTests/Tests/DecodedTests.swift
@@ -56,6 +56,26 @@ class DecodedTests: XCTestCase {
     }
   }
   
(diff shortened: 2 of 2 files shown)


tonyd256:
Do you think we need to support the case where they want to customize the error message that is returned in a `Custom` error case? As in, have a second parameter that's a closure or something with the an `ErrorType` parameter maybe.

paulyoung:
Yeah, I think so. We could use `error.description` if that parameter isn't provided.

paulyoung:
> have a second parameter that's a closure or something with the an `ErrorType` parameter maybe

I'm not sure what you mean by that exactly. How about just have that parameter be a `String`?

tonyd256:
So I'm thinking that someone might want to customize the error message based on the error returned. I could see 3 implementations for this function `(() throws -> T) -> Decoded<T>`, `(() throws -> T, String) -> Decoded<T>`, and `(() throws -> T, ErrorType -> String) -> Decoded<T>` ... does that make sense? I'm not sure that this is something we want to do or even if it's useful. 

@gfontenot thoughts?

tonyd256:
Those 3 implementations might combine to 1 implementation with the use of `@autoclosure` and a default.

gfontenot:
I feel like we should lean on the custom error type for now. If this gets enough use to warrant the other use cases, we'll notice.

tonyd256:
I'm cool with that :+1:

tonyd256:
merged here: 220a492102d2e9c080b0b2890da7a22be1bb8b00

PR #265, opened 2015-10-08
Don't sign builds for the simulator

gfontenot:
There's no real reason to sign builds for the simulator. This actually
breaks builds with Carthage on CI systems that are only trying to build
for the simulator anyway.


tonyd256:
Does this problem also effect the Mac target? If not then merge away!

gfontenot:
not sure about the watchOS target here :/ I'm unclear on if I should be using the iOS simulator setting or the watchOS simulator setting

PR #271, opened 2015-10-20
Mention Curry dependency in README

### LINKED ISSUE #270: Explicitly mention Curry is external in README
The README code sample uses `curry` but the `curry` functions do not come with Argo. They are part of another library, [Curry](https://github.com/thoughtbot/Curry). We should reference this in the README to clarify that this will require an extra dependency.

relevant: #269

gilesvangruisen:
fixes #270

Perhaps it would be worthwhile to show an example of a curried `create` method in case people don't want to add an additional dependency?

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ can then link against `Argo.framework` for your application target.
 
 ## Usage tl;dr:
 
+Please note: the example below requires an additional, external module named [Curry](https://github.com/thoughtbot/Curry) which lets us use the `curry` function to curry `User.init`.
+
 ```swift
 import Argo
 import Curry


tonyd256:
I was thinking it might get more visibility being either before the example or right after the install guide

gilesvangruisen:
Yeah I had it above the example initially. Happy to move it back there.

README.md
```diff
@@ -100,6 +100,8 @@ if let j: AnyObject = json {
 
 For more information, see the [Documentation](Documentation/)
 
+Please note: the above example requires an additional, external module named [Curry](https://github.com/thoughtbot/Curry) to automatically curry `User.init`.
```
tonyd256: also it's not automatic so we should say "to use the `curry` function to curry `User.init`." or something like that

gilesvangruisen: True, "automatic" was a poor choice of wording. I'll make those changes once I'm back at my computer :+1:

gilesvangruisen:
@tonyd256 thoughts on providing an example of a curried `create` method?

README.md
```diff
@@ -100,6 +102,7 @@ if let j: AnyObject = json {
 
 For more information, see the [Documentation](Documentation/)
 
+
```
tonyd256: extra line?

tonyd256:
I think an example of a curried create is something we want to have in our documentation. Then we could have a link in that blurb that pointed to the appropriate docs

gilesvangruisen:
Good call, I think that makes more sense than adding clutter to the README

PR #277, opened 2015-10-27
Update errors for RawRepresentable types.

### LINKED ISSUE #275: Better error message for enums/RawRepresentable types
I currently see `TypeMismatch(Expected .Some(Provider), got .None)` because of an invalid value coming from the server and ultimately being passed to `Provider(rawValue: ...)`

It would be useful to know what the value is without needing to go and look at the server response.

paulyoung:
This aims to close #275.

Using my `Provider` example:

```swift
enum Provider {
  case Facebook = "facebook"
  case GitHub = "github"
  case Twitter = "twitter"
}
```

**Before:**
```
TypeMismatch(Expected .Some(Provider), got .None)
```

**After:**
```
TypeMismatch(Expected Provider, got String(google))
```

### DIFF
--- a/Argo/Extensions/RawRepresentable.swift
+++ b/Argo/Extensions/RawRepresentable.swift
@@ -1,7 +1,7 @@
 public extension Decodable where Self.DecodedType == Self, Self: RawRepresentable, Self.RawValue == String {
   static func decode(json: JSON) -> Decoded<Self> {
     switch json {
-    case let .String(s): return self.init(rawValue: s).map(pure) ?? .typeMismatch("String", actual: json)
+    case let .String(s): return self.init(rawValue: s).map(pure) ?? .typeMismatch("\(self)", actual: json)
     default: return .typeMismatch("String", actual: json)
     }
   }
@@ -10,7 +10,7 @@ public extension Decodable where Self.DecodedType == Self, Self: RawRepresentabl
 public extension Decodable where Self.DecodedType == Self, Self: RawRepresentable, Self.RawValue == Int {
   static func decode(json: JSON) -> Decoded<Self> {
     switch json {
-    case let .Number(n): return self.init(rawValue: n as Int).map(pure) ?? .typeMismatch("Int", actual: json)
+    case let .Number(n): return self.init(rawValue: n as Int).map(pure) ?? .typeMismatch("\(self)", actual: json)
     default: return .typeMismatch("Int", actual: json)
     }
   }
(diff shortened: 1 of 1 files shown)


tonyd256:
I'm not sure that `After` is better. Would you mind posting the code that would throw the error as well?

I'd like to maintain type information while showing the user what the original value was. I liked your original proposal better with this format: `TypeMismatch(Expected X, got Y from/with Z)`.

paulyoung:
> Would you mind posting the code that would throw the error as well?

In this case, the code that causes the error is essentially `self.init(rawValue: "google")` where `"google"` is the value in the JSON but doesn't match a raw value in the enum.

@tonyd256 is that what you meant?

tonyd256:
Gotcha. Looking through Argo, this is the only place we use `fromOptional`. And because of that, this is the only place we're not showing the `actual` json but instead just focus on the Optional. I think this is a good change!

paulyoung:
How about this?

`TypeMismatch(Expected rawValue for Provider, got String(google))`

tonyd256:
Would it be too much to print out all options?  Something like:
`TypeMismatch(Expected one of [facebook, github, twitter], got google)`
This would probably require some reflection but is the error explicitness worth it?
Also, do you think this would get too big in large enum cases?

paulyoung:
> Would it be too much to print out all options?

Personally, I'm happy to go and look at the enum if I'm not sure of the expected raw values. I think the fact that obtaining the actual raw value required making the request again was the real pain point for me.

> Also, do you think this would get too big in large enum cases?

I have an enum of ISO 3166-1 country codes that I definitely wouldn't want printed out!

> This would probably require some reflection

I'm not too familiar with reflection in Swift so can't speak to any performance implications or safety. I've avoided it so far but think the other points I mentioned mean it might not be worth it.

tonyd256:
Ya, I figured that might be a bit much. I like your second suggestion best I think: `TypeMismatch(Expected rawValue for Provider, got String(google))`

@gfontenot any thoughts here?

gfontenot:
Yeah, I had the same thought re: using crazy reflection stuff to print out the expected values, but I agree that it's probably overkill and not worth the time/complexity. I think that the proposed solution (`Expected rawValue for Provider, got String(google)`) is a solid improvement without going crazy.

paulyoung:
Updated!

gfontenot:
Awesome. Merged in as da89baf.

PR #281, opened 2015-10-30
Fix CI with Circle and Carthage

gfontenot:
- Install Carthage from a pkg
 - Only install Carthage dependencies if needed
 - Update bin/test to work without xctool
 - Use Xcode 7
 - Cache the Carthage directory

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,3 +1,10 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
-xctool -project Argo.xcodeproj -scheme Argo-Mac -sdk macosx clean test
+set -o pipefail
+
+xcrun xcodebuild \
+  -workspace Argo.xcworkspace \
+  -scheme Argo-Mac \
+  test \
+  | xcpretty --color
+
--- a/circle.yml
+++ b/circle.yml
@@ -1,5 +1,11 @@
 machine:
-  environment:
-    XCODE_SCHEME: Argo-Mac
   xcode:
-    version: 6.3.1
+    version: "7.1"
+
+dependencies:
+  override:
+    - git submodule update --init --recursive
+
+test:
+  override:
+    - bin/test
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,9 @@
 machine:
   xcode:
     version: "7.1"
+  environment:
+    XCODE_SCHEME: nonce
+    XCODE_WORKSPACE: nonce.xcworkspace
 
 dependencies:
   override:


bin/install-carthage
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+version=0.9.4
+
+curl -L -O "https://github.com/Carthage/Carthage/releases/download/$version/Carthage.pkg"
```
paulyoung: I wonder if this is subject to rate limits.

I ran into something with Travis and Carthage binaries. I know that's using a different mechanism but it seemed to be checking the IP address.

gfontenot: Hmm, good call. I haven't _seen_ that problem (using this same technique on a client app right now), but I'll keep an eye on it.

bin/carthage-bootstrap-if-needed
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+if ! diff Cartfile.resolved Carthage/Cartfile.resolved &>/dev/null; then
+  carthage bootstrap 
+  cp Cartfile.resolved Carthage
+fi
```
paulyoung: How does `Carthage/Cartfile.resolved` exist?

paulyoung: I guess this isn't just for CI. I _think_ the intention must be to commit `Carthage/Cartfile.resolved` at some point?

paulyoung: Or maybe it's to do with caching that directory?

gfontenot: We're copying the `Cartfile.resolved` into the `Carthage` directory on line 5. That directory is then cached by Circle (specified in circle.ml)

Although, it occurs to me that since we're using git submodules, we don't actually need to do _any_of this, so I'm going to remove it.

paulyoung: I see. I was using git submodules directly on CI but I've been installing Carthage just so I can use the same script as locally.

paulyoung: Do you know what command CircleCI uses for git submodules? On Travis it didn't do `--recursive` or something like that so I had to disable the built in support and then call it manually.

gfontenot: I don't. I was going to try the default, and tweak it from there if it didn't work.

gfontenot:
There's something wrong on Circle's end here. I don't know what's going on, but it seems like it isn't respecting our circle.yml or something. I'm contacting their support about it now.

hyperspacemark:
How does this diff cause Carthage to be installed via package?

gfontenot:
It actually doesn't anymore. If I removed it in favor of just using git submodules. If you want to see a better diff with Carthage, check out [this PR for Runes](https://github.com/thoughtbot/Runes/pull/55).

Although, Carthage is actually pre-installed on Circle now. :tada:

gfontenot:
Kachow. Build is green. Merging.

PR #282, opened 2015-10-30
Enable Hound

gfontenot:


gfontenot:
@jakecraige This should have triggered a violation, right?

Argo/Types/JSON.swift
```diff
@@ -54,3 +54,5 @@ public func == (lhs: JSON, rhs: JSON) -> Bool {
   default: return false
   }
 }
+
+let x :String = "Break"
```
gfontenot: :tada:

PR #286, opened 2015-11-06
Add establishment json decoding stress test.

drhaynes:
This PR is not for merging, but to give an overview of Argo and outline performance.

Argo is a type-safe, declarative, functional json decoding library inspired by functional parsing techniques developed by Haskell nerds.

We're using a json file containing 5000 establishment objects, which are essentially POIs with some additional metadata attached. Example object:

```
{  
   "id":2859510,
   "fhrsid":507136,
   "name":"196",
   "address":"Mill Road, Cambridge, Cambridgeshire",
   "postcode":"CB1 3NF",
   "rating":5,
   "rating_date":"2012-12-03",
   "lat":52.197346,
   "long":0.145034,
   "created_at":"2015-01-17T07:28:47.385Z",
   "updated_at":"2015-01-17T07:28:47.385Z",
   "validated":false
}
```

We create a simple model struct to mirror this:

```
struct Establishment {
  let id: Int
  let fhrsid: Int
  let name: String?
  let address: String?
  let postcode: String?
  let rating: Int
  let ratingDate: String?
  let latitude: Float?
  let longitude: Float?
  let createdDate: String
  let updatedDate: String
  let validated: Bool
}
```

Once the data model is defined, you setup the mapping from model object properties to json fields declaratively by implementing Argo's `Decodable` protocol (which comprises entirely of the `decode` method).

For our Establishment struct, that ends up looking like this:

```
extension Establishment: Decodable {
    static func decode(j: JSON) -> Decoded<Establishment> {
        curry(self.init)
            <^> j <| "id"
            <*> j <| "fhrsid"
            <*> j <|? "name"
            <*> j <|? "address"
            <*> j <|? "postcode"
            <*> j <| "rating"
            <*> j <|? "rating_date"
            <*> j <|? "lat"
            <*> j <|? "long"
            <*> j <| "created_at"
            <*> j <| "updated_at"
            <*> j <| "validated"
    }
}
```

The operator overloading is gross and impenetrable, here's a cheat sheet:

`<^>` - map (This needs to be first - we map the init function to the rest of the parsing declaration).
`<*>` - apply (subsequent fields are 'defined' using this. Think of it like connecting all the individual parsing definition functions together to make the overall definition).
`<|` - parse value (property/field this corresponds to must not be an optional)
`<|?` - parse optional value (property/field this corresponds to must be optional)

Once you've got a blob of json you can then parse it into model objects like so:

`let establishments: [Establishment]? = json.flatMap(decode)`

If any objects fail to parse, we'll get nil and they'll be binned in the `flatMap`. If the whole thing fails to parse, we'll end up with a nil array. No exceptions, no crash-crashy.

Performance is outlined below:

Average time to parse 5000 records (includes initial loading from disk to NSData and conversion to json using NSJSONSerialization.JSONObjectWithData): 47.5s
Peak memory usage: 61mb

SwiftyJSON will be tested with the same methodology.


drhaynes:
Apologies, meant to open this on our team's own fork. It's not at all intended to be merged.

gfontenot:
Hahaha. No apologies needed. Let us know how we stack up. :wink:

PR #291, opened 2015-11-08
Remove `CustomStringConvertible` constraint on `typeMismatch` function

dapenggao:
According to Apple's [documentation](https://developer.apple.com/library/prerelease/ios/documentation/Swift/Reference/Swift_CustomStringConvertible_Protocol/index.html) of `CustomStringConvertible`: 
> `String(instance)` will work for an `instance` of _any_ type, returning its `description` if the `instance` happens to be `CustomStringConvertible`. Using `CustomStringConvertible` as a generic constraint, or accessing a conforming type's `description` directly, is therefore discouraged.

Hence we can unify the `typeMismatch` function so that it accepts _any_ types.

### DIFF
--- a/Argo/Types/Decoded/Decoded.swift
+++ b/Argo/Types/Decoded/Decoded.swift
@@ -5,14 +5,14 @@ public enum Decoded<T> {
   public var value: T? {
     switch self {
     case let .Success(value): return value
-    default: return .None
+    case .Failure: return .None
     }
   }
-  
+
   public var error: DecodeError? {
     switch self {
+    case .Success: return .None
     case let .Failure(error): return error
-    default: return .None
     }
   }
 }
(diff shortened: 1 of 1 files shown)


tonyd256:
I like this :+1: ... snuck in some other clean up stuff but I like those too.

gfontenot:
This is great. Can you move that explanation you provided into body of the git commit message and then I'm happy to merge this in?

gfontenot:
If you could also rebase on master while your at it, that'll fix CI for this PR.

dapenggao:
Done here. Added explanation and CI passes.

dapenggao:
Any updates on this?

gfontenot:
I'm not sure what happened here, but it looks like your fork is no longer available? That's preventing me from being able to merge it.

dapenggao:
Sorry about that. I'll open another PR.

Issue #290, opened 2015-11-07
Question: How do I use functions for Dictionary?


S-Shimotori:
I built Argo.framework with Xcode 7C46t and Carthage 0.9.4 and set it in ```Linked Frameworks and Libraries``` ,
but I can't call ```<^>``` in https://github.com/thoughtbot/Argo/blob/master/Argo/Extensions/Dictionary.swift

I can use the functions in the other files.

```swift
let a: Decoded<Int> = .missingKey("") //OK
```

however I failed in calling ```+``` and ```map``` in ```Extensions/Dictionary.swift``` .

```swift
let b = ["one": 1] + ["two": 2] //Binary operator '+' cannot applied
let c = ["three": 3].map //Xcode suggests only this closure: ((String, Int)) throws -> T
```
 
Do I forget some settings or use these functions mistakenly?

dapenggao:
These function are not exposed as `public` in Argo. That's why you can't use them.

S-Shimotori:
Can I set ```public``` ? Or I will make my own functions if you have the reason they need ```internal``` .

I want to get ```[String: Decoded<Decodable>]``` , not ```Decoded<[String: Decodable]>``` .

gfontenot:
As noted in the proposed PR, I think it's outside Argo's scope to provide these. We kept them internal because we need them for our own implementation, but polluting the public scope with them seems like a bad idea. If you need these functions, I'd recommend either writing them yourself or finding a library that provides these functions for you.

PR #292, opened 2015-11-09
Add `public` Dictionary functions

S-Shimotori:
#290

Can Argo make these functions ```public```?


gfontenot:
Sorry, been neglectful this weekend.

I don't think it's within Argo's scope to add these functions publicly. If you need them for yourself, I recommend adding them yourself, or finding an external library that provides them for you.

Thanks for the PR!

Issue #296, opened 2015-11-19
Handling [String: Decodable] in Swift 2


mikebuss:
How would one handle decoding [String: Decodable] in Swift 2? In Swift 1, the following code worked:

```swift
struct LocationHours {
    let type: String
    let title: String
    let subtitle: String?
    let comment: String?
    let times: [String: LocationTime]
}

extension LocationHours: Decodable {
    static func create(type: String)(title: String)(subtitle: String?)(comment: String?)(times: [String: LocationTime]) -> LocationHours {
        return LocationHours(type: type, title: title, subtitle: subtitle, comment: comment, times: times)
    }
    
    static func decode(j: JSON) -> Decoded<LocationHours> {
        return LocationHours.create
            <^> j <| "type"
            <*> j <| "title"
            <*> j <|? "subtitle"
            <*> j <|? "comment"
            <*> j <| "times" >>- decodeObject
   
    }
}
```

But in Swift 2 (Argo 2.2.0), I get the following error:

```Use of unresolved identifier `decodeObject````

I wasn't able to find this in the documentation. Apologies if I missed it!

**Comment by gfontenot** [comment]
`decodeObject` has been removed in favor of an extension on `Dictionary`. This should be equivalent:

```swift
<*> j <| "times" >>- [String: LocationTime].decode
```

you _might_ need parens.

mikebuss:
Thanks for the quick response. I was able to get it working with the parenthesis:

```swift
    static func decode(j: JSON) -> Decoded<LocationHours> {
        return LocationHours.create
            <^> j <| "type"
            <*> j <| "title"
            <*> j <|? "subtitle"
            <*> j <|? "comment"
            <*> (j <| "times" >>- [String: LocationTime].decode)
        
    }
```

PR #297, opened 2015-11-19
Add free functions for decoding collections

gfontenot:
In Argo 2 we removed the free functions `decodeArray` and `decodeObject`
in favor of adding this functionality to the types themselves. I believe
this was a mistake and would like to reintroduce them to the public API.

My reasoning is this:

Adding `decode` to these types does _not_ make these types conform to
`Decodable`. That's confusing to me (and I'd imagine, other people). The
result of this is that you can't just decode directly to these types,
but instead have to use a special function. In the case of `Array`
that's not a huge deal since we expose an operator (`<||`) for this, but
for decoding objects, it means that users have to specify the actual
type that they'd like to decode. This is weird, since it's the _only_
place in the public API that we don't let users lean on the type system
to do the hard work for them.

I think it makes the most sense to reintroduce these free functions
without actually removing the existing protocol extensions. This will
improve backwards compatibility with Argo 1 and improve what I see as a
deficiency in the current API. In addition, if Swift ever gets the
ability to do conditional protocol conformance, leaving the existing
protocol extensions will make it easy to support that as well.

See #296 for an example of where we've made a user's code worse with the new
API.

### DIFF
--- a/Argo/Types/StandardTypes.swift
+++ b/Argo/Types/StandardTypes.swift
@@ -72,6 +72,10 @@ public extension CollectionType where Generator.Element: Decodable, Generator.El
   }
 }
 
+public func decodeArray<T: Decodable where T.DecodedType == T>(j: JSON) -> Decoded<[T]> {
+  return [T].decode(j)
+}
+
 public extension DictionaryLiteralConvertible where Value: Decodable, Value == Value.DecodedType {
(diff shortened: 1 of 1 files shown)


tonyd256:
:+1:

Chiliec:
:+1:

PR #298, opened 2015-11-19
[RFC]: Remove Optional versions of `<|` operators

gfontenot:
The addition of overloads for `<|` and friends that return optional
values sounded like a great idea but I think they have ended up being
awkward in practice.

The problem that they solve is, frankly, minimal. It helps when decoding
from a root key, but that's about it.

On the other hand, introducing an entirely new set of operators adds
overhead and introduces possible ambiguity for the compiler as well as
introduces cognitive overhead for the user. I'm not convinced that some
of the issues we've seen around expression complexity and poor error
messaging aren't due to the introduction of these (rarely used)
operators.

The _real_ solution would be for Swift to support Monads and then we
could make these completely generic, but I'm not holding my breath
there.

I'm not planning on merging this yet, just want to open a discussion about it.


tonyd256:
:+1:

gfontenot:
Another datapoint that this is probably the best long-term move, the following code is completely ambiguous:

```swift
let foos: [Foo] = (json <|| "foos") ?? []
```

If we only return `Decoded` here, this isn't an issue anymore.

paulyoung:
I've yet to use the `Optional` versions and don't anticipate needing them often, so no objections here :+1:

gfontenot:
This has been open for a _really_ long time with no objections, so I'm going to move this into master in preparation for 3.0

jshier:
Too late, but I used it fairly often to manually decode keys from a root JSON when the representation was really bad and I didn't want to deal with handling the `Decoded` versions:
```swift
public static func decode(j: JSON) -> Decoded<NSDate> {
    guard let timeString: String = j <| "time",
        periodString: TimePeriod = j <| "timeSection" else {
        return .customError("Expected time String, got \(j).")
    }
        
    let timePeriodString = "\(timeString) \(periodString)"
        
   return .fromOptional(DateFormatter.scheduleTimeFormatter.dateFromString(timePeriodString))
}
```
In any event I think I'll move the optional operators into my own code.

gfontenot:
yep, that's really the only reasonable usecase that we knew of. However, you could replicate the old behavior by adding `.value` so it seemed like it wasn't adding all that much:

```swift
// before
let foo: Foo? = j <| "rootKey"

// after
let foo: Foo? = (j <| "rootKey").value
```

gfontenot:
> wasn't adding all that much

^^ not quite right. It's definitely useful, but it had an adverse affect on the complexity for the compiler, as well as caused headaches when the type was fairly ambiguous. Removing `<|` for `Optional` simplifies those things at the cost of losing this one feature, which seemed like a reasonable tradeoff.

jshier:
Yeah, I'll just use `.value` instead of using the custom operators. Thanks.

Issue #300, opened 2015-11-30
String parsing for Int64?


nvh:
As Javascript is [not really](http://www.2ality.com/2012/07/large-integers.html) equipped to handle large integers, and it's required to use strings for anything above 53-bits, wouldn't it be logical if the decode function of `Int64` automatically parses strings?

tonyd256:
that makes sense to me. Would you be able to open a PR for it?

nvh:
I tried to that this afternoon, but cloning Argo and running `bin/setup` crashed carthage. That was, however, before I read [this](https://github.com/Carthage/Carthage/issues/812). After running `submodule init` and `submodule update`, all works fine again. Anyway, I will try again tomorrow :)

gfontenot:
Merged in the fix from #301 as e7680fc

PR #301, opened 2015-11-30
Added parsing from String for decoding Int64

nvh:
Implementation of solution for https://github.com/thoughtbot/Argo/issues/300

### DIFF
--- a/Argo/Types/StandardTypes.swift
+++ b/Argo/Types/StandardTypes.swift
@@ -22,6 +22,12 @@ extension Int64: Decodable {
   public static func decode(j: JSON) -> Decoded<Int64> {
     switch j {
     case let .Number(n): return pure(n.longLongValue)
+    case let .String(s):
+      if let i = Int64(s) {
+        return pure(i)
+      } else {
+        return .typeMismatch("Int64", actual: s)
+      }
     default: return .typeMismatch("Int64", actual: j)
     }
   }
--- a/Argo/Types/StandardTypes.swift
+++ b/Argo/Types/StandardTypes.swift
@@ -26,7 +26,7 @@ extension Int64: Decodable {
       if let i = Int64(s) {
         return pure(i)
       } else {
-        return .typeMismatch("Int64", actual: s)
+        fallthrough
       }
     default: return .typeMismatch("Int64", actual: j)
     }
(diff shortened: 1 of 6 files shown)


Argo/Types/StandardTypes.swift
```diff
@@ -22,6 +22,12 @@ extension Int64: Decodable {
   public static func decode(j: JSON) -> Decoded<Int64> {
     switch j {
     case let .Number(n): return pure(n.longLongValue)
+    case let .String(s):
+      if let i = Int64(s) {
+        return pure(i)
+      } else {
+        return .typeMismatch("Int64", actual: s)
```
gfontenot: Possible to use `fallthrough` here?

gfontenot:
I think this is great. Solid idea keeping `Number` support while adding `String` support.

Argo/Types/StandardTypes.swift
nvh: Good point, fixed in e08fed7a288533ce6c80f9a0c80fde6e15ed6dcb

nvh:
I was trying to get it in 1 line, but there is no way to do something like
```swift
if case let .String(s) = j, let i = Int64(s)
```
in a switch statement, right?

Argo/Types/StandardTypes.swift
```diff
@@ -22,6 +22,12 @@ extension Int64: Decodable {
   public static func decode(j: JSON) -> Decoded<Int64> {
     switch j {
     case let .Number(n): return pure(n.longLongValue)
+    case let .String(s):
+      if let i = Int64(s) {
+        return pure(i)
+      } else {
```
tonyd256: shouldn't need the `else` ... it will return if it succeeds, stoping execution.

tonyd256: could possibly use a guard?

tonyd256:
I like this. only comment is to condense that case somehow.

Argo/Types/StandardTypes.swift
gfontenot: +1

```swift
guard let i = Int64(s) else { fallthrough }
return pure(i)
```

should work nicely.

nvh: Fixed in 2e9047c

gfontenot:
Fantastic. Do you mind squashing and rebasing and then I'll merge this in?

gfontenot:
Rebased and merged as e7680fc

Thanks!

PR #302, opened 2015-12-05
Remove `CustomStringConvertible` constraint on `typeMismatch` function

dapenggao:
According to Apple's documentation of `CustomStringConvertible`: `String(instance)` will work for an `instance` of _any_ type, returning its `description` if the `instance` happens to be `CustomStringConvertible`. Using `CustomStringConvertible` as a generic constraint, or accessing a conforming type's `description` directly, is therefore discouraged.

Hence we can unify the `typeMismatch` function so that it accepts _any_ types.


gfontenot:
(noting that this supersedes #291)

Thanks!

Issue #255, opened 2015-10-02
Flesh out inline documentation


gfontenot:
Our monadic operators have full documentation, but that's about it. We should really document the entire public API inline so that it can be picked up by Xcode, as well as Jazzy/CocoaDocs

gfontenot:
List of things that need to be documented (AFAIK):

- [ ] `RawRepresentable`
- [ ] `catDecoded`
- [ ] `decode`
- [ ] `decode(_: rootKey:)`
- [ ] `flatReduce`
- [ ] `sequence`
- [ ] `<|` for `Decoded<T>`
- [ ] `<||` for `Decoded<[T]>`
- [ ] `<|` for `Optional<T>` (unless removed)
- [ ] `<||` for `Optional[T]>` (unless removed
- [ ] `Decodable`
- [ ] `DecodeError`
- [ ] `Decoded`
- [ ] `JSON`
- [ ] `Alternative` functions for `Decoded`
- [ ] `Applicative` functions for `Decoded`
- [ ] `Functor` functions for `Decoded`
- [ ] `Monad` functions for `Decoded`
- [ ] `FailureCoalescing` for `Decoded`
- [ ] Default `T.decode` functions for standard types

There's a first-run at the documentation for all of these available in #309. I'd rather break that up into per-file PRs to make reviewing/rebasing/merging easier.

gfontenot:
All open PRs about this have been merged 🎉

I think this is safe to close now.

PR #309, opened 2015-12-13
Add inline documentation for the public interface

gfontenot:
This adds a _bunch_ of inline documentation. I tried to cover the entire
public interface, which is extensive. I think there is still room for
improvement, but this is a _huge_ leap forward IMO, and probably gets us
80-90 percent of the way there.

### DIFF
--- a/Argo/Functions/catDecoded.swift
+++ b/Argo/Functions/catDecoded.swift
@@ -1,9 +1,32 @@
+/**
+  Remove `Failure` values from an array and unwrap `Success`es
+
+  This will iterate through the array of `Decoded` values and ensure that each
+  is `Success`. If it is, it will unwrap the contained `T` value. If it is
+  `Failure`, it will be removed from the array.
+
+  - parameter xs: An array of `Decoded` values
+
+  - returns: An array of unwrapped values of type `T`
+*/
 public func catDecoded<T>(xs: [Decoded<T>]) -> [T] {
   return xs.reduce([]) { accum, elem in
     elem.map { accum + [$0] } ?? accum
   }
 }
 
+/**
+  Remove `Failure` values from a dictionary and unwrap `Success`es
+
+  This will iterate through the dictionary of `Decoded` values and ensure that
+  each is `Success`. If it is, it will unwrap the contained `T` value and assign
+  it back to the original key. If it is `Failure`, it will be removed from the
+  dictionary, along with the original key.
+
+  - parameter xs: A dictionary of `Decoded` values assigned to `String` keys
+
+  - returns: A dictionary of unwrapped values of type `T` assigned to `String` keys
+*/
 public func catDecoded<T>(xs: [String: Decoded<T>]) -> [String: T] {
   return xs.reduce([:]) { accum, elem in
     elem.1.map { accum + [elem.0: $0] } ?? accum
--- a/Argo/Functions/decode.swift
+++ b/Argo/Functions/decode.swift
@@ -1,35 +1,215 @@
-// MARK: decode root objects and arrays
+/**
+  Decode a raw JSON value into an optional `Decodable` value
 
+  This function takes the output from `NSJSONSerialization` and attempts to
+  decode it into a `Decodable` value, returning an `Optional`. This works based
+  on the type you ask for.
+
+  For example, the following code attempts to decode an optional `String`,
+  because that's what we have explicitly stated is the return type:
+
+  ```
+  do {
+    let object = try NSJSONSerialization.JSONObjectWithData(data, options: nil)
+    let str: String? = decode(object)
+  } catch {
+    // handle error
+  }
+  ```
+
+  - parameter object: The output from `NSJSONSerialization`
+  - returns: An `Optional` value of type `T` where `T` is `Decodable`
+*/
 public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) -> T? {
   return decode(object).value
 }
 
+/**
+  Decode a raw JSON value into an optional array of `Decodable` values
+
+  This function takes the output from `NSJSONSerialization` and attempts to
+  decode it into an array of `Decodable` values, returning an `Optional`. This
+  works based on the type you ask for.
+
+  For example, the following code attempts to decode an optional array of
+  `String`s, because that's what we have explicitly stated is the return type:
+
+  ```
+  do {
+    let object = try NSJSONSerialization.JSONObjectWithData(data, options: nil)
+    let str: [String]? = decode(object)
+  } catch {
+    // handle error
+  }
+  ```
+
+  - parameter object: The output from `NSJSONSerialization`
+  - returns: An `Optional` array of values of type `T` where `T` is `Decodable`
+*/
 public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) -> [T]? {
   return decode(object).value
 }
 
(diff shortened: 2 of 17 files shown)


gfontenot:
Interesting output from `cloc`:

```
$ cloc Argo
      22 text files.
classified 22 files
      22 unique files.                              
       1 file ignored.

http://cloc.sourceforge.net v 1.64  T=0.11 s (188.9 files/s, 11944.9 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Swift                           20            255            669            395
C/C++ Header                     1              3              3              3
-------------------------------------------------------------------------------
SUM:                            21            258            672            398
-------------------------------------------------------------------------------
```

gfontenot:
cc @thoughtbot/copy-editing and @thoughtbot/ios

gfontenot:
Fixes #255

Argo/Functions/catDecoded.swift
```diff
@@ -1,9 +1,32 @@
+/**
+  Remove `Failure` values from an array and unwrap `Success`es
+
+  This will iterate through the array of `Decoded` values and ensure that each
+  is `Success`. If it is, it will unwrap the contained `T` value. If it is
+  `Failure`, it will be removed from the array.
+
+  - parameter xs: An array of `Decoded` values
```
gabebw: Maybe `Decoded<T>` instead of `Decoded`?

Argo/Functions/catDecoded.swift
```diff
@@ -1,9 +1,32 @@
...
 
+/**
+  Remove `Failure` values from a dictionary and unwrap `Success`es
+
+  This will iterate through the dictionary of `Decoded` values and ensure that
+  each is `Success`. If it is, it will unwrap the contained `T` value and assign
+  it back to the original key. If it is `Failure`, it will be removed from the
+  dictionary, along with the original key.
+
+  - parameter xs: A dictionary of `Decoded` values assigned to `String` keys
```
gabebw: Same here -- `Decoded<T>`?

Argo/Functions/catDecoded.swift
```diff
@@ -1,9 +1,32 @@
+/**
+  Remove `Failure` values from an array and unwrap `Success`es
```
gabebw: Is the standard to not have a period at the end of this first line of documentation?

Argo/Functions/decode.swift
```diff
@@ -1,35 +1,215 @@
-// MARK: decode root objects and arrays
+/**
+  Decode a raw JSON value into an optional `Decodable` value
```
gabebw: uppercase Optional?

Argo/Functions/decode.swift
```diff
@@ -1,35 +1,215 @@
...
+  do {
+    let object = try NSJSONSerialization.JSONObjectWithData(data, options: nil)
+    let str: String? = decode(object)
+  } catch {
+    // handle error
+  }
+  ```
+
+  - parameter object: The output from `NSJSONSerialization`
+  - returns: An `Optional` value of type `T` where `T` is `Decodable`
```
gabebw: Maybe: 

> An `Optional<T>` value where `T` is `Decodable`

Argo/Functions/decode.swift
gabebw: If this should be uppercase, there are a bunch of other lowercase "optional"s that should also be changed.

Argo/Functions/decode.swift
```diff
@@ -1,35 +1,215 @@
...
+*/
 public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) -> T? {
   return decode(object).value
 }
 
+/**
+  Decode a raw JSON value into an optional array of `Decodable` values
```
sidraval: How do you feel about using "Parse" instead of Decode at the beginning of this sentence (or really any sentence that mentions the `Decodable` type)? Maybe it's just me but I have to double take on every sentence that uses Decode twice like this.

gfontenot: This is probably a bigger discussion about our language for the project. I _still_ don't think we've come up with a agreed upon word for the `<|` family of operators. I have been using `decode` there, but it _is_ a really overloaded term in this lib. I'm also not sure that I totally understand what the difference is between decoding and parsing.

I also think it'd be weird to say that the `decode` function "Parses a raw JSON value into an optional array of `Decodable` values"

sidraval: @gfontenot do you think this is a valuable place to have that discussion? I think getting language we're happy with in the documentation is a good start.

How do you feel about "Transforms a raw JSON value into an optional array of Decodable values" instead of "Parses"? That way it's about data type transformation, and we leave the loaded words parse/decode out of it?

gfontenot: `Transform` might be better.

I don't mind having the discussion here. Seems reasonable. I just wanted to make sure it was noted that this was a bigger conversation than the wording of this one bit of documentation.

Argo/Functions/catDecoded.swift
gfontenot: I never know how to communicate this properly, honestly. Adding in `<T>` here seems like it _might_ be confusing

gfontenot: that was supposed to be more of a question than a statement

Argo/Functions/catDecoded.swift
gfontenot: I have no idea

Argo/Functions/catDecoded.swift
tonyd256: I think using the full type signature is ok ... `Decoded<T>` ... it also makes more sense as you read on

Argo/Functions/decode.swift
```diff
@@ -1,35 +1,215 @@
-// MARK: decode root objects and arrays
+/**
+  Decode a raw JSON value into an optional `Decodable` value
 
+  This function takes the output from `NSJSONSerialization` and attempts to
```
tonyd256: would it be too confusing to say it takes `AnyObject` instead ... then later say "for example the output of NSJSONSerial..."?

Argo/Functions/decode.swift
```diff
@@ -1,35 +1,215 @@
...
+  ```
+  do {
+    let object = try NSJSONSerialization.JSONObjectWithData(data, options: nil)
+    let str: String? = decode(object)
+  } catch {
+    // handle error
+  }
+  ```
+
+  - parameter object: The output from `NSJSONSerialization`
```
tonyd256: here too ... actually this is `AnyObject`

Argo/Functions/decode.swift
```diff
@@ -1,35 +1,215 @@
...
+  - returns: An `Optional` array of values of type `T` where `T` is `Decodable`
+*/
 public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) -> [T]? {
   return decode(object).value
 }
 
+/**
+  Decode a raw JSON value into a decoded `Decodable` value
```
tonyd256: if you're speaking of our `JSON` type then should this be in backtics?

gfontenot:
This is way out of date, and of an unmanageable size. I'm going to close this in favor of breaking this up into smaller more digestible chunks.

PR #310, opened 2015-12-16
Add alternative named instance method on Decoded

### LINKED ISSUE #307: Add `alternative` as a instance method for `Decoded`
All of our other monadic operators (`<^>`, `<*>`, `>>-`) have named instance methods that contain the actual implementation (`map`, `apply`, `flatMap`). I believe that having these named instance methods helps with education and discoverability (not to mention providing concrete names to wacky symbols). However, this isn't true for `<|>`.

We should add `Decoded.alternative` as a named instance method, using the current implementation for `<|>`. Then `<|>` should be defined as `return lhs.alternative(rhs)`

DorianListens:
Fixes #307, brings alternative inline with other monadic operators in
Argo.

### DIFF
--- a/Argo/Types/Decoded/Alternative.swift
+++ b/Argo/Types/Decoded/Alternative.swift
@@ -13,8 +13,14 @@
 infix operator <|> { associativity left precedence 140 }
 
 public func <|> <T>(lhs: Decoded<T>, @autoclosure rhs: () -> Decoded<T>) -> Decoded<T> {
-  if case .Success = lhs {
-    return lhs
+  return lhs.or(rhs)
+}
+
+public extension Decoded {
+  func or(@autoclosure other: () -> Decoded<T>) -> Decoded<T> {
+    switch self {
+      case .Success: return self
+      case .Failure: return other()
+    }
   }
-  return rhs()
 }
--- a/ArgoTests/Tests/DecodedTests.swift
+++ b/ArgoTests/Tests/DecodedTests.swift
@@ -116,6 +116,30 @@ class DecodedTests: XCTestCase {
       XCTFail("Unexpected Error Occurred")
     }
   }
+
+  func testDecodedOrWithSuccess() {
+    let successUser: Decoded<User> = decode(JSONFromFile("user_with_email")!)
+    let failedUser: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
+
+    let result = successUser.or(failedUser)
+
+    switch result {
+    case .Success: XCTAssert(result.description == successUser.description)
+    default: XCTFail("Unexpected Case Occurred")
+    }
+  }
+
+  func testDecodedOrWithError() {
+    let successUser: Decoded<User> = decode(JSONFromFile("user_with_email")!)
+    let failedUser: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
+
+    let result = failedUser.or(successUser)
+
+    switch result {
+    case .Success: XCTAssert(result.description == successUser.description)
+    default: XCTFail("Unexpected Case Occurred")
+    }
+  }
 }
 
 private struct Dummy: Decodable {


Argo/Types/Decoded/Alternative.swift
```diff
@@ -13,8 +13,15 @@
...
 
 public func <|> <T>(lhs: Decoded<T>, rhs: Decoded<T>) -> Decoded<T> {
-  if case .Success = lhs {
-    return lhs
+  return lhs.alternative(rhs)
+}
+
+public extension Decoded {
+  func alternative<T>(other: Decoded<T>) -> Decoded<T> {
+    if case .Success = self {
```
gfontenot: Can we change this to using a `switch`?

```swift
switch self {
case .Success: return self
case .Failure: return other
}
```

You shouldn't need any force casting, and I think it's a bit more explicit about the intent.

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -116,6 +116,35 @@ class DecodedTests: XCTestCase {
       XCTFail("Unexpected Error Occurred")
     }
   }
+
+  func testDecodedAlternativeWithSuccess_ReturnsSelf() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_email")!)
+    let user_error: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
```
gfontenot: This doesn't appear to be used here

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -116,6 +116,35 @@ class DecodedTests: XCTestCase {
...
+    let alternative = user.alternative(user_error)
```
gfontenot: Naming nit: can we name this `otherUser`?

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -116,6 +116,35 @@ class DecodedTests: XCTestCase {
...
+
+    XCTAssertEqual(user.value?.id, alternative.value?.id)
+  }
+
+  func testDecodedAlternativeWithError_ReturnsOther() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_email")!)
+    let user_error: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
```
gfontenot: Can we name this `failedUser`?

ArgoTests/Tests/DecodedTests.swift
gfontenot: Oh, wait, I see. Cann this be called `failedUser`?

ArgoTests/Tests/DecodedTests.swift
gfontenot: Lets call this `result`

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -116,6 +116,35 @@ class DecodedTests: XCTestCase {
...
+    let alternative = user_error.alternative(user)
```
gfontenot: can we name this `result`?

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -116,6 +116,35 @@ class DecodedTests: XCTestCase {
...
+
+    XCTAssertEqual(user.value?.id, alternative.value?.id)
```
gfontenot: I believe this should be able to be `XCTAssert(result == user)`

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -116,6 +116,35 @@ class DecodedTests: XCTestCase {
       XCTFail("Unexpected Error Occurred")
     }
   }
+
+  func testDecodedAlternativeWithSuccess_ReturnsSelf() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_email")!)
+    let user_error: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
+    let alternative = user.alternative(user_error)
+
+    XCTAssertEqual(user.value?.id, alternative.value?.id)
```
gfontenot: I believe this should be able to be `XCTAsserc(result == user)

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -116,6 +116,35 @@ class DecodedTests: XCTestCase {
...
+
+  func testDecodedAlternativeWithError_ReturnsOther() {
+    let user: Decoded<User> = decode(JSONFromFile("user_with_email")!)
+    let user_error: Decoded<User> = decode(JSONFromFile("user_with_bad_type")!)
+    let alternative = user_error.alternative(user)
+
+    XCTAssertEqual(user.value?.id, alternative.value?.id)
+  }
+
+  func testDecodedAlternativeWithTwoErrors_ReturnsOther() {
```
gfontenot: This test is redundant

gfontenot:
One comment on the implementation, and I have some high level thoughts on the tests:

 - We shouldn't be using `_` anywhere in our names, tests or variables
 - Can we be more explicit with out names? `result` for the result of calling `alternative`, `successUser` and `failedUser` for the successful and failing decoding operations of `User`

gfontenot:
fwiw, [antitypical/Result](https://github.com/antitypical/Result/blob/master/Result/Result.swift#L72) calls this function `recoverWith`, which is probably a better name than `alternative`. I wonder if we should follow suit?

cc @thoughtbot/ios

Argo/Types/Decoded/Alternative.swift
DorianListens: I'd like to do this without force casting, but on my machines, this removing it (using either the if or the switch implementation) results in this excellent error message:  cannot convert return expression of type 'Decoded<T>' to return type 'Decoded<T>'

DorianListens:
Thanks for the comments. I think I should be able to clean this up today or tomorrow!

Argo/Types/Decoded/Alternative.swift
gfontenot: Ah, I see. That's because you're actually declaring a new type `T`. Subtle bug. Make this change on line 20:

```diff
-  func alternative<T>(other: Decoded<T>) -> Decoded<T> {
+  func alternative(other: Decoded<T>) -> Decoded<T> {
```

and it should all work

DorianListens: Nice catch!

ArgoTests/Tests/DecodedTests.swift
DorianListens: Unfortunately: "Binary operator '==' cannot be applied to two 'Decoded<User>' operands"

gfontenot: Huh. I would have expected us to have `==` for `Decoded` if the constrained type was equatable. We should probably do that.

DorianListens:
Okay, back from vacation mode. I've responded to all your feedback, and think this is ready to merge.
Should I rebase + squash, or is it fine to merge as is?

gfontenot:
My only remaining concern is the name of the method. It was recently suggested that we use `or`, which seems nice especially in use.

```swift
foo.or(pure("bar"))
```

DorianListens:
Fair enough. I like "or" better than "recoverWith", for what it's worth.

gfontenot:
@DorianListens If you can rename the method from `alternative` to `or` and rebase/squash, I'll be happy to merge this in!

DorianListens:
Just finishing this up now. Should I go as far as renaming the file to "Or" as well?

gfontenot:
> Should I go as far as renaming the file to "Or" as well?
> 


No, it's still part of the Alternative typeclass, so the file name should stay the same.

DorianListens:
Okay, all done. Interesting fact: since the bin/test "upgrade", I can no longer successfully execute that script. Gotta love working in Apple's ecosystem.

DorianListens:
Like, all the tests pass on all the platforms when I run them individually from Xcode, and I can even make the script succeed if I start it, and then run the tests for the various platforms from within Xcode.

But if I just run it on the command line, it fails to properly boot the simulator - it launches whichever was the last simulator used from Xcode, but never gets past a black screen.

Possibly because I'm running Xcode 7.2? Any ideas?

paulyoung:
@DorianListens are your command line tools also set to 7.2?

paulyoung:
By the way, I had to add `-destination` for iOS to get `xcodebuild` to work with projects on 7.2

https://github.com/the-grid/Portal/blob/master/script/test#L4

DorianListens:
Just now, the App Store offered me an update to the command line tools, and now everything works,

Thanks for the suggestion, @paulyoung 
Sorry for the unrelated noise in this PR.

DorianListens:
I've rebased this on Master again, accounting for the autoclosure. Should be good to go!

gfontenot:
:tada: Thanks!

Issue #295, opened 2015-11-15
Performance issues


RuiAAPeres:
I found that on a 4S device, my data model was taking almost 12 seconds, to be parsed. So I did a small experiment with the following:

```
import Foundation
import Argo
import Curry
import Runes

struct Person {
    
    let firstName: String
    let secondName: String?
    let age: Int
    let hoobies: [String]
    let description: String
}

extension Person: Decodable {
    
    static func decode(j: JSON) -> Decoded<Person> {
        return curry(Person.init)
            <^> j <| "firstName"
            <*> j <|? "secondName"
            <*> j <| "age"
            <*> j <|| "hobbies"
            <*> j <| "description"
    }
}
```

This is not my original model, but a far simpler one. So I used a testing file with 100k lines of JSON with a total of 9130 objects. The original one has more than 330k lines and far deeper hierarchy and complexity (number of members). 

I wasn't sure which would the best approach so I used the ones I saw in the wild (the measurements are in release mode on a MacbookPro 2015):

<img width="928" alt="screen shot 2015-11-15 at 21 26 40" src="https://cloud.githubusercontent.com/assets/949286/11171109/89243156-8bdf-11e5-8cdf-3eda215cd3b4.png">

Then I did the parsing "by hand": 

<img width="933" alt="screen shot 2015-11-15 at 21 28 09" src="https://cloud.githubusercontent.com/assets/949286/11171111/9c241154-8bdf-11e5-8e0c-f85b52315772.png">

Obviously, the second version has a lot more boilerplate, it's not as easy to maintain or extend, but in this case all those niceties are not worth with such timings. A demo project can be found [here](https://www.dropbox.com/s/053wr6vib1ssimn/TestingStuff.zip?dl=0).

RuiAAPeres:
Btw, this is the memory footprint while measuring `test_measure_Argo_1` and `test_measure_By_Hand`. The red circle is from the later. 
<img width="902" alt="screen shot 2015-11-15 at 22 26 02" src="https://cloud.githubusercontent.com/assets/949286/11171436/670bc1e8-8be8-11e5-9fa7-53012a7d965b.png">

jshier:
First, 330k lines is a ridiculous amount of data for a mobile app. Can you not page in just bits of the data at a time?

In any case, I ran an equivalent test case through Instruments. The time profiler for:
```swift
let people: Decoded<[Person]> = [Person].decode(objectJSON)
```
breaks down like this:
```
[...]

Issue #185, opened 2015-07-31
Make decoding lazy


coenert:
Now we have parse the whole JSON value into the `JSON` type, but then you have a lot of overhead when you have a lot of properties that you don't use. I tried this with a type `DecodableValue` that has a helper method to [cast](https://github.com/coenert/Argo/blob/master/Argo/Types/DecodableValue.swift#L19-L27) the data of the `DecodableValue` to the type that you want. `DecodableValue` replaces JSON and also saves a lot of looping and `switch`es.

gfontenot:
This is something we're definitely looking to do, especially since we're starting to use Argo with more complex JSON structures, like those adhering to [JSON:API](http://jsonapi.org).

PR #313, opened 2016-01-04
[RFC] Change Argo to lazy parse json

tonyd256:
To lazily parse json we change the `JSON` type to not be recursive.
Arrays and Objects no longer parse their values when the json is parsed
into a `JSON` value. Instead, we call parse on the values just before
they are decoded. The major pitfall here is that there is duplicate
parsing since previously parsed json will have to be parsed again if
used again.

Looking at the performance tests: Before this change, parsing took about
137ms and decoding took 37ms totalling 174ms. After this change, parsing
took 0ms (this makes sense) and decoding took 100ms totalling 100ms. So
we see that decoding took longer since we moved the parsing into it, but
we see the overall time is shorter and even shorter than just parsing
was.

This is an experimental attempt to implement #185 and might hopefully solve or improved #295

### DIFF
--- a/Argo/Types/JSON.swift
+++ b/Argo/Types/JSON.swift
@@ -1,8 +1,8 @@
 import Foundation
 
 public enum JSON {
-  case Object(() -> [Swift.String: JSON])
-  case Array(() -> [JSON])
+  case Object([Swift.String: LazyJSON])
+  case Array([LazyJSON])
   case String(Swift.String)
   case Number(NSNumber)
   case Null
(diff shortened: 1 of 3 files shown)


jferris:
Interesting. I'm not sure how the performance behaves, but another option is to use closures/currying to build a tree of functions which return `JSON` values instead of building a tree of `AnyObject` references. That would preserve the closed type system you previously had with the recursive type, whereas `AnyObject` technically allows invalid structures to compile.

This would also allow you to layer in a cache for already-parsed values, because you could swap the parsing function with a constant function after the first invocation.

paulyoung:
I like @jferris' idea!

tonyd256:
@jferris I like the theory, I'm not sure I get the implementation ... might have to talk this one out with you in slack.

gfontenot:
@tonyd256 what's the status here? Is this still something we want to look into? Did @jferris have more ideas on how to implement this in a reasonable way?

tonyd256:
Unfortunately, I don't really like any of the ways I've tried to implement laziness above. Maybe there is something to look into with `LazySequence` but I haven't given it much time.

gfontenot:
I'm going to close this PR for now, then. I think we should still look into this as a performance enhancement in the future, but it sounds like this isn't the way forward.

PR #314, opened 2016-01-05
Add a Boolean case to JSON

gfontenot:
Previously, we were overloading `JSON.Number` to include boolean values.
This was done because we believed that there was no way to determine if
an `NSNumber` returned from `NSJSONSerialization` was actually a boolean
value or not. Turns out, this was incorrect.

Under the covers, `NSNumber` is toll-free bridged with `CFNumberRef`.
And `CFNumberRef` has a specific implementation for boolean values
(`CFBooleanRef`). So we can actually get the type ID for `CFBooleanRef`
and just check it against the type ID for `self` and if that returns
`true`, then we have a boolean value and can handle it accordingly.

Fixes #314


paulyoung:
@gfontenot were you planning on adding any tests for `isBool`?

gfontenot:
@paulyoung ....

yes of course I was

gfontenot:
@paulyoung oh, actually we can't. We test under release settings (because of concerns with swift optimizations) so we can't do `@testable import` and we don't want to export that extension publicly.

_obviously_ I knew that when I wrote this and then just forgot.

I think that the type tests validate that this works as intended.

paulyoung:
@gfontenot I see :wink: 

If it makes sense, I think it would be good to have test coverage for decoding things like the following:

-  `{ "bool": 1 }`
-  `{ "number": 1 }`
-  `{ "bool": 2 }`

paulyoung:
FWIW I don't use `@testable` and instead test the public API :+1:

tonyd256:
:+1: love this

PR #315, opened 2016-01-05
Improve bin/test

gfontenot:
Previously we were _only_ running tests on CI against the mac target.
This seems short sighted. Instead, lets make sure we can build for all
of our supported platforms, running the tests on the platforms that
support it.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -8,3 +8,24 @@ xcrun xcodebuild \
   test \
   | xcpretty --color
 
+xcrun xcodebuild \
+  -workspace Argo.xcworkspace \
+  -scheme Argo-iOS \
+  -destination "platform=iOS Simulator,name=iPhone 6" \
+  test \
+  | xcpretty --color
+
+xcrun xcodebuild \
+  -workspace Argo.xcworkspace \
+  -scheme Argo-watchOS \
+  -destination "generic/platform=watchOS" \
+  CODE_SIGNING_REQUIRED=NO \
+  CODE_SIGN_IDENTITY="" \
+  | xcpretty --color
+
+xcrun xcodebuild \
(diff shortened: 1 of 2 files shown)


gfontenot:
Ok, now this is working.

Alternatively, it might be nice to run all of the tests and then run `carthage build --no-skip-current` to make sure that all of the exposed schemes build.

@thoughtbot/ios thoughts?

circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   xcode:
-    version: "7.0"
+    version: "7.1"
```
tonyd256: no 7.2 yet :(

bin/test
```diff
@@ -8,3 +8,24 @@ xcrun xcodebuild \
...
+  -destination "platform=iOS Simulator,name=iPhone 6" \
+  test \
+  | xcpretty --color
+
+xcrun xcodebuild \
+  -workspace Argo.xcworkspace \
+  -scheme Argo-watchOS \
+  -destination "generic/platform=watchOS" \
+  CODE_SIGNING_REQUIRED=NO \
+  CODE_SIGN_IDENTITY="" \
```
tonyd256: so is the watch one special since we can't run tests on it. This is just making sure it builds?

tonyd256:
:+1:

circle.yml
gfontenot: Nope.

bin/test
gfontenot: Right. There's no such thing as a test bundle that will run again a watchOS framework. Fun stuff. Making sure it builds seems like a reasonable alternative for now. Basically, I want to make sure CI catches people not adding files to all of the frameworks.

PR #317, opened 2016-01-11
Remove reduce usage for better performance

jshier:
Argo’s performance is quadratic over the size of the collection being decoded, due to the use of reduce ([AirspeedVelocity article](http://airspeedvelocity.net/2015/08/03/arrays-linked-lists-and-performance/)). This PR removes most usages of `reduce` in exchange for a `for` loop of the same functionality. This improves the performance of decoding an array of over 9000 simple models on an iPhone 4S by ~2600% (76.4s to 2.9s). See #295 for my analysis and evolution of this code.

### DIFF
--- a/Argo/Functions/catDecoded.swift
+++ b/Argo/Functions/catDecoded.swift
@@ -15,8 +15,8 @@ public func catDecoded<T>(xs: [Decoded<T>]) -> [T] {
 public func catDecoded<T>(xs: [String: Decoded<T>]) -> [String: T] {
   var accum = Dictionary<String, T>(minimumCapacity: xs.count)
 
-  for (key, decodedValue) in xs {
-    switch decodedValue {
+  for (key, x) in xs {
+    switch x {
     case let .Success(value): accum[key] = value
     case .Failure: continue
     }
--- a/Argo/Functions/catDecoded.swift
+++ b/Argo/Functions/catDecoded.swift
@@ -1,11 +1,26 @@
 public func catDecoded<T>(xs: [Decoded<T>]) -> [T] {
-  return xs.reduce([]) { accum, elem in
-    elem.map { accum + [$0] } ?? accum
+  var accum: [T] = []
+  accum.reserveCapacity(xs.count)
+  
+  for x in xs {
+    switch x {
+    case let .Success(value): accum.append(value)
+    case .Failure: continue
+    }
   }
+  
+  return accum
 }
 
 public func catDecoded<T>(xs: [String: Decoded<T>]) -> [String: T] {
-  return xs.reduce([:]) { accum, elem in
-    elem.1.map { accum + [elem.0: $0] } ?? accum
+  var accum = Dictionary<String, T>(minimumCapacity: xs.count)
+  
+  for (key, decodedValue) in xs {
+    switch decodedValue {
+    case let .Success(value): accum[key] = value
+    case .Failure: continue
+    }
   }
+  
+  return accum
 }
--- a/Argo/Functions/catDecoded.swift
+++ b/Argo/Functions/catDecoded.swift
@@ -1,26 +1,26 @@
 public func catDecoded<T>(xs: [Decoded<T>]) -> [T] {
   var accum: [T] = []
   accum.reserveCapacity(xs.count)
-  
+
   for x in xs {
     switch x {
     case let .Success(value): accum.append(value)
     case .Failure: continue
     }
   }
-  
+
   return accum
 }
 
 public func catDecoded<T>(xs: [String: Decoded<T>]) -> [String: T] {
   var accum = Dictionary<String, T>(minimumCapacity: xs.count)
-  
+
   for (key, decodedValue) in xs {
     switch decodedValue {
     case let .Success(value): accum[key] = value
     case .Failure: continue
     }
   }
-  
+
   return accum
 }
--- a/Argo/Functions/sequence.swift
+++ b/Argo/Functions/sequence.swift
@@ -16,7 +16,7 @@ public func sequence<T>(xs: [Decoded<T>]) -> Decoded<[T]> {
 
 public func sequence<T>(xs: [String: Decoded<T>]) -> Decoded<[String: T]> {
   var accum: [String: T] = [:]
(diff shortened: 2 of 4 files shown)


Argo/Functions/catDecoded.swift
```diff
@@ -1,11 +1,26 @@
...
+
+  return accum
 }
 
 public func catDecoded<T>(xs: [String: Decoded<T>]) -> [String: T] {
-  return xs.reduce([:]) { accum, elem in
-    elem.1.map { accum + [elem.0: $0] } ?? accum
+  var accum = Dictionary<String, T>(minimumCapacity: xs.count)
+
+  for (key, decodedValue) in xs {
```
tonyd256: you use `x` above for `decodedValue`, want to use `x` here too?

Argo/Functions/sequence.swift
```diff
@@ -1,11 +1,30 @@
...
+
+  return pure(accum)
 }
 
 public func sequence<T>(xs: [String: Decoded<T>]) -> Decoded<[String: T]> {
-  return xs.reduce(pure([:])) { accum, elem in
-    curry(+) <^> accum <*> ({ [elem.0: $0] } <^> elem.1)
+  var accum: [String: T] = [:]
+
+  for (key, value) in xs {
```
tonyd256: can we use `x

Argo/Functions/sequence.swift
```diff
@@ -1,11 +1,30 @@
 public func sequence<T>(xs: [Decoded<T>]) -> Decoded<[T]> {
-  return xs.reduce(pure([])) { accum, elem in
-    curry(+) <^> accum <*> ({ [$0] } <^> elem)
+  var accum: [T] = []
+  accum.reserveCapacity(xs.count)
+
+  for elem in xs {
```
tonyd256: and here, `x`

tonyd256:
Thanks for doing this. The performance increases are amazing! Really sad to see the functional code go away, but if `for` loops are more efficient then so be it.

My only comments are about consistency in the variable naming but otherwise this is good.

jshier:
@tonyd256 Done. I also updated some layout, and added a `Dictionary` minimum capacity I missed.

tonyd256:
@jshier I was curious if you would be able to write a test similar to our PerformanceTests.swift (or perhaps just update that test) to use whatever JSON you're using in those test results you posted in #295. I think it's important that we track this performance in the future and asking you to run the test every commit/release isn't scalable :smiley:

jshier:
I was just thinking about that. I'll put together a reduced case of perhaps 500 - 1000 items, so it runs in a reasonable amount of time, just to make sure.

tonyd256:
Great! If it's simple copy-paste of the JSON to get it back up to the one you're testing here then that would be fine.

jshier:
@tonyd256 There is a large test file already in the project (`big_data.json`). However, it's not correct for the `TestModel` type and so fails decoding, making the performance test it's used for meaningless. I've built a new `big_data.json` that is correct, and big enough to see a difference, though the impact is lower when running on the Mac. How much time is acceptable for the tests?

jshier:
For instance, with a new test file with 10,000 `TestModel`s, tests take 37s to run on my 3.5GHz i7 iMac, with my optimizations. Same file with the old sequence takes 123s. Time taken to decode decreased from about 10s to 1.2s.

tonyd256:
Haha, no wonder the performance tests ran so fast on my machine. The `big_data.json` file must not have been updated along with the `types.json` file. We should definitely add a check for success to those performance tests. Would you mind doing that as part of the test changes you're already making?

And I think that's fine for timing.

jshier:
@tonyd256 I've added my updated `big_data.json` file, though I'm unsure how we can check for the success of the decode without affecting performance. I suppose I could just assert that the count of the array is 10,000? I don't know what impact that assertion will have from inside the `measureBlock`, as I can't find a way to capture the result of one of the runs outside the block.

Also, I've tried to rebase / squash, but my local squash doesn't seem to get synced up. Any idea what I'm doing wrong?

tonyd256:
Thanks @jshier for updating this! As for the assertion that it can parse, I think it would be find to have a separate test that doesn't use a performance block but does assert that there are 10000 objects. 

About the rebase, not sure what happened but it seems you have a wrong commit as the first commit of this PR (908aa92d2c5657efec8d571f3af6afe6271b68a0). If you rebase off master perhaps just deleting that commit will be helpful. Then you should be able to force push you local rebase/squash to you're remove fork and then github should pull that in here right? Is that the part that's not happening?

jshier:
I'll add that test.

I think the force push is what I was missing.

liscio:
To add an extra data point to the speed improvement here, I ran this branch against a rather small 32k+ line JSON file, and a related test case went from 4.850s down to 0.771s on a MBP13 with a 2.6GHz Core i5.

Does this sort of enhancement _have_ to wait 'til the 3.0 milestone? Performance issues like this scare folks off (if the custom operators didn't already), and it doesn't appear to affect the existing API. Folks that already shipped against the 2.x API could pick up the performance gain before having to deal with any API change. Thoughts?

gfontenot:
@liscio Oh wow. Yeah, we (@tonyd256 and I) should talk about trying to get this performance fix out ASAP before moving towards a 3.0 release.

gfontenot:
@tonyd256 @jshier what's the status here? Is this ready to merge? I'd like to get a 2.3 release out this week that includes this improvement.

jshier:
@gfontenot This is good to go, I just have to figure out how to squash it down appropriately.

gfontenot:
@jshier I'm happy to do that for you if you don't want to deal with it. I'd probably merge the first 4 into one commit and the last 3 into another. I'd also rebase to get rid of the merge commit.

jshier:
That would be fine. Thanks.

gfontenot:
Merged this as 416349d and 255196c

Thanks so much for this, @jshier!

gfontenot:
Argo 2.3 is now available and includes this change

liscio:
:+1: :+1:

PR #319, opened 2016-01-15
Add leftwise flatMap operator for Decoded

### LINKED ISSUE #308: Add left-wise `flatMap` operator to `Decoded`
Right now we're only providing `>>-` for `Decoded`. It'd probably be useful to provide `-<<` as well.

This operator is already defined by Runes, so we'd just need to add the implementation and documentation.

gfontenot:
This version of the operator was left out for some reason. This wasn't
intentional, so we're just correcting the oversight.

Fixes #308


tonyd256:
:+1:

PR #320, opened 2016-01-15
Use an autoclosure for the rhs of <|>

### LINKED ISSUE #289: Should Alternative use an autoclosure?
`??`, which is arguably similar to Alternative, uses an autoclosure for it's right hand value. Does it then make sense to do the same for Alternative?

gfontenot:
The nil coalescer operator already uses an autoclosure for the rhs
argument, so it makes sense to follow suit here.

Fixes #289


tonyd256:
:+1:

gilesvangruisen:
I like this change :+1:

PR #324, opened 2016-01-26
Enable bitcode for all targets except Mac

mbrandonw:
I was getting this error while using Carthage to build a tvOS dependency which had Argo as a dependency:

```
ld: bitcode bundle could not be generated because '/Carthage/Build/tvOS/Argo.framework/Argo' 
was built without full bitcode. All frameworks and dylibs for bitcode must be generated from 
Xcode Archive or Install build for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

** BUILD FAILED **
```

Pretty sure we want bitcode to be enabled for all targets except Mac. This indeed how it is in Curry too.

Does this sound about right?

Strangely, when I build the dependency directly I do not get this error. Only when Argo was a dependency of a dependency.

### DIFF
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -1259,7 +1259,6 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				APPLICATION_EXTENSION_API_ONLY = YES;
-				BITCODE_GENERATION_MODE = marker;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=watchsimulator*]" = "";
 				DEBUG_INFORMATION_FORMAT = dwarf;
@@ -1285,7 +1284,6 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				APPLICATION_EXTENSION_API_ONLY = YES;
-				BITCODE_GENERATION_MODE = bitcode;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=watchsimulator*]" = "";
 				COPY_PHASE_STRIP = NO;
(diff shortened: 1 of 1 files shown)


mbrandonw:
Has anyone else run into this? I might only be seeing this because Argo is a dependency of a dependency :/

gfontenot:
@mbrandonw What version of Carthage are you using? I was under the impression that this was handled by them at this point.

mbrandonw:
@gfontenot 0.11.0, though I just this moment upgraded to 0.12.0. I could try that if you think it would be different.

gfontenot:
I honestly have no idea. I can probably take a look on Friday if nobody beats me to it.

gfontenot:
@mbrandonw Have you still been having this problem?

mbrandonw:
I'm not, but only cause I abandoned carthage and just use submodules now! :P 

If I'm the only one reporting this there's a good chance I just had something else wrong with my setup. I'll close this...

inamiy:
Hi, I might have encountered the similar situation.

In my [RAC project](https://github.com/inamiy/ReactiveCocoaCatalog), I'm using Carthage for checkout only, and Argo v4.1.1 + Runes v4.0.1 xcodeprojs are imported to xcworkspace.

And when I did Release + Device + Build, I got a bitcode error as following:

```
ld: bitcode bundle could not be generated because '/Users/inamiy/Library/Developer/Xcode/DerivedData/ReactiveCocoaCatalog-awpxvsrxtsodsgbwirfwfpizaiqb/Build/Products/Release-iphoneos/Runes.framework/Runes' was built without full bitcode. All frameworks and dylibs for bitcode must be generated from Xcode Archive or Install build for architecture armv7
```

After looking through Argo and Runes settings, I found the following fields were missing:

1. `BITCODE_GENERATION_MODE` is missing in all targets of Runes, so above error occurs.
2. `BITCODE_GENERATION_MODE` is missing in tvOS target of Argo.
3. `BITCODE_GENERATION_MODE` is (unnecessarily) set in macOS target of Argo. 

I think Argo and Runes needs to have the same settings.

PR #332, opened 2016-03-16
Refactor loops to use `for case` & `if case` syntax

alskipp:
The catDecoded functions ignore the `Failure` case, so we don’t need to
pattern match for `Failure`.


gfontenot:
Is there a reasoning behind this change? Or is it purely cosmetic?

alskipp:
Mainly cosmetic, but thought it was worthwhile to remove a few lines of code.

gfontenot:
Personally, I actually like the full `switch` here. It makes it explicit that there are only two options.

Thanks so much for working on this! However, I don't think we're going to make this change.

PR #333, opened 2016-03-22
Update to Swift 2.2

leoMehlig:
A few code changes to build without warnings in Xcode 7.3

### DIFF
--- a/Argo/Types/Decoded/Decoded.swift
+++ b/Argo/Types/Decoded/Decoded.swift
@@ -22,7 +22,8 @@ public extension Decoded {
     switch x {
     case let .Success(value): return .Success(.Some(value))
     case .Failure(.MissingKey): return .Success(.None)
-    case let .Failure(.TypeMismatch(x)): return .Failure(.TypeMismatch(x))
+    case let .Failure(.TypeMismatch(expected, actual)):
+      return .Failure(.TypeMismatch(expected: expected, actual: actual))
     case let .Failure(.Custom(x)): return .Failure(.Custom(x))
     }
   }
--- a/Argo/Extensions/Dictionary.swift
+++ b/Argo/Extensions/Dictionary.swift
@@ -1,10 +1,11 @@
 // pure merge for Dictionaries
-func + <T, U>(var lhs: [T: U], rhs: [T: U]) -> [T: U] {
+func + <T, U>(lhs: [T: U], rhs: [T: U]) -> [T: U] {
+  var merged = lhs
   for (key, val) in rhs {
-    lhs[key] = val
+    merged[key] = val
   }
 
-  return lhs
+  return merged
 }
 
 extension Dictionary {
--- a/Argo/Types/Decodable.swift
+++ b/Argo/Types/Decodable.swift
@@ -1,4 +1,4 @@
 public protocol Decodable {
-  typealias DecodedType = Self
+  associatedtype DecodedType = Self
   static func decode(json: JSON) -> Decoded<DecodedType>
 }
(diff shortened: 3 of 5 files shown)


Argo/Types/Decoded/Decoded.swift
```diff
@@ -22,7 +22,8 @@ public extension Decoded {
     switch x {
     case let .Success(value): return .Success(.Some(value))
     case .Failure(.MissingKey): return .Success(.None)
-    case let .Failure(.TypeMismatch(x)): return .Failure(.TypeMismatch(x))
+    case let .Failure(.TypeMismatch(x)):
```
tonyd256: what about changing this to 
```
.Failure(.TypeMismatch(expected, actual)):
  return .Failure(.TypeMismatch(expected: expected, actual: actual))
```

tonyd256:
One comment but otherwise looks good

leoMehlig:
Changed like suggested in the note

mbrandonw:
Should this also update the Curry dependency to 2.0.0 since that also fixes a bunch of compiler warnings?

leoMehlig:
@mbrandonw I can run `carthage update` and commit the changes made by Carthage.
I can also rebase the two commits to one if that is wanted.

gfontenot:
We've been discussing this internally and think it probably makes sense to hold off on merging this till we ship for Swift 3. Given that the source will compile fine under 2.1 and 2.2 in its current state, but merging this update would mean needing to jump to Argo 3.0 because of the breaking change (then jumping again to 4.0 for Swift 3 support shortly thereafter), I feel like we should live with the warnings for now.

We'll merge this into master once we have a more concrete plan on when we're going to try to ship Swift 3 support.

jshier:
Semantic versioning doesn't require major version bumps because of increased tooling requirements, only public API changes, which these are not (AFAICT). Alamofire recently dealt with this as well, [here's the explanation](https://github.com/Alamofire/Alamofire/blob/master/CHANGELOG.md#upgrade-notes).

mbrandonw:
ReactiveCocoa did the same for their 4.0.1 ~> 4.1 release, which added Swift 2.2 compatibility.

I'm using Argo in an app that gets quite a bit of development, and it's becoming cumbersome to deal with 60+ warnings in Xcode. Seems worth the departure from semver.

gfontenot:
I think that releasing code that fails to compile on older toolchains _does_ constitute a breaking change and would require a major version bump. I think that the fact that AlamoFire needs a disclaimer to explain what's going on only serves to prove that point. Additionally, bumping the major version for source compatibility changes would be consistent with how we've handled this in the past. Argo 2.0 was released to support Swift 2.0.

THAT BEING SAID: I also couldn't care less about our version number climbing rapidly or making full major jumps every time Swift changes source compatibility. There's way too much ceremony around versioning, and I have no interest in adding to it. So if there's a demand (which it sounds like there is), I'm happy to release 3.0 with Swift 2.2 support this week. It'd probably be a branch off of the 2.3 release instead of including everything currently on `master`.

I will say that I'm curious about the number of warnings mentioned by @mbrandonw: Are you using CocoaPods/Carthage for integration? Or are you using Git Submodules? The former two both allow you to ignore our warnings in your source, which is what drove the initial decision to avoid these changes till absolutely necessary, so I'm interested to hear about your experiences.

mbrandonw:
Would def love to see this go to 3.0 and not care about the actual numbers involved in versions :clap: 

@gfontenot alas I'm just using plain submodules :( I could of course fork and fix on my own if that's the minority situation. mostly just wanna know what the plan is so i can do something about the warnings.

gfontenot:
> mostly just wanna know what the plan is so i can do something about the warnings.

Noted. I think it makes sense to release this sooner rather than later. I'm going to commune with @tonyd256 and come up with a release plan, but I think I'll sneak some of these (breaking) API simplifications into a 3.0 release as well, and make 4.0 a speed/swift 3.0 release. I'll try to get this released this week.

mbrandonw:
thanks @gfontenot!

PR #336, opened 2016-04-02
Use an initializer instead of a factory for JSON

gfontenot:
Previously, we were using a static func here because we needed to recursively
pass it during initial parsing. However, since we can now pass the initializer,
I think it makes sense to simplify the API by directly using an initializer.

### DIFF
--- a/Argo/Types/JSON.swift
+++ b/Argo/Types/JSON.swift
@@ -20,13 +20,23 @@ public extension JSON {
 
     - returns: An instance of the `JSON` tree structure
   */
-  static func parse(json: AnyObject) -> JSON {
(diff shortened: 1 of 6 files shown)


tonyd256:
:+1: great improvement!

Argo/Types/JSON.swift
```diff
@@ -50,3 +60,12 @@ public func == (lhs: JSON, rhs: JSON) -> Bool {
   default: return false
   }
 }
+
+/// MARK: Deprecations
+
+extension JSON {
+  @available(*, deprecated=3.0, renamed="init")
```
tonyd256: oh I like this!

PR #338, opened 2016-04-02
Add inline documentation for `RawRepresentble`

gfontenot:
We provide a couple of default implementations for if your enum type is
`RawRepresentable` using either `String` or `Int` values. This adds
simple documentation for those implementations.

### DIFF
--- a/Argo/Extensions/RawRepresentable.swift
+++ b/Argo/Extensions/RawRepresentable.swift
@@ -1,3 +1,7 @@
+/**
+  Default implementation of `Decodable` for `RawRepresentable` types using
+  `String` as the raw value.
+*/
 public extension Decodable where Self.DecodedType == Self, Self: RawRepresentable, Self.RawValue == String {
   static func decode(json: JSON) -> Decoded<Self> {
     switch json {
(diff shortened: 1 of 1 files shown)


Argo/Extensions/RawRepresentable.swift
```diff
@@ -1,3 +1,7 @@
+/**
+  Default implementation of `Decodable` for `RawRepresentable` types using
+  `String` as the raw value.
```
edwardloveall: Since this is a very automatic function that results is less code, it might be good to write out an example:

If you have a type `State`:

```swift
enum State: String {
  case Unstarted = "unstarted"
  case Started = "started"
  case Completed = "completed"
}
```

which means the rawValue of each case is represented by a string (`"unstarted"`, `"started"`, `"completed"`), this allows Argo to automatically decode the data into types:

```swift
// code the developer has to write, if anything, that makes that work
```

edwardloveall:
One comment for you to consider, otherwise it's fine.

tonyd256:
:+1: Looks good. I like the idea of an example or perhaps a link to the docs where the example is.

Argo/Extensions/RawRepresentable.swift
gfontenot: I kinda think this might be better for full documentation, as opposed to inline documentation. I do think we could add more documentation around this though.

PR #339, opened 2016-04-02
Improve the API for root key decoding

gfontenot:
I think the original reason for ordering the arguments as `rootKey`,
`object` was to be able to curry the function and partially apply the root
key (as seen in the tests). IMO, this is a weird optimization and let to a
kinda funky API. I think it probably makes more sense to accept that
currying in this way isn't reasonable, and instead order the arguments so
that our `decode` functions have a consistent pattern.

I've also re-ordered these functions for pedantic reasons.

### DIFF
--- a/Argo/Functions/decode.swift
+++ b/Argo/Functions/decode.swift
@@ -125,18 +125,18 @@ public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject) -> 
   ```
   do {
     let object = try NSJSONSerialization.JSONObjectWithData(data, options: nil)
-    let str: Decoded<String> = decodeWithRootKey("value", object)
+    let str: Decoded<String> = decode(object, rootKey: "value")
   } catch {
     // handle error
   }
   ```
 
-  - parameter rootKey: The root key that contains the object to decode
   - parameter object: The `AnyObject` instance to attempt to decode
+  - parameter rootKey: The root key that contains the object to decode
 
   - returns: A `Decoded<T>` value where `T` is `Decodable`
 */
-public func decodeWithRootKey<T: Decodable where T == T.DecodedType>(rootKey: String, _ object: AnyObject) -> Decoded<T> {
+public func decode<T: Decodable where T == T.DecodedType>(object: AnyObject, rootKey: String) -> Decoded<T> {
   return JSON(object) <| rootKey
 }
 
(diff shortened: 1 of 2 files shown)


gfontenot:
All of this came to me while writing documentations for these functions. Changing the type from `AnyObject` to `[String: AnyObject]` seems like it's a big difference, but it felt weird needing to explain that via documentation when we could make the expectations explicit via the type itself.

tonyd256:
:+1: This makes sense

tonyd256:
@gfontenot after thinking about it ... what happens if the `AnyObject` returned from `NSJSONSerialization` isn't a dictionary? Would it crash when trying to pass it into this function or would the user have to cast it before? Anyways, what I'm getting at is that if the `AnyObject` is not a dictionary then you'll get a `Decoded` failure here. With this change, I'm unsure what will happen.

gfontenot:
@tonyd256 The user would need to cast the `AnyObject` to `[String: AnyObject]`. This is a bit of a change from the way we're doing things now, but it's actually much more explicit and I think it's a reasonable tradeoff.

Before:

```swift
let anyObj = NSJSONSerialization.whatevs
let obj: Foo? = anyObj.map { decodeWithRootKey("key", $0) }
```

After:

```swift
let anyObj = NSJSONSerialization.whatevs() as? [String: AnyObject]
let obj: Foo? = anyObj.map { decode($0, rootKey: "key") }
```

tonyd256:
Hmm, I'm a bit torn on this one. It is taking that conversion out of our hands but I think it could potentially hurt the user as it would now silently fail. In that example, it will silently fail regardless because it is requesting the optional form, but if the user requested the `Decoded` form, then the error would be silent. Keeping it `AnyObject` means we can return an error that says "you have an array but we expected an object" which could be helpful. thoughts?

tonyd256:
:+1:

gfontenot:
Had to re-do a lot of this because this didn't mesh up with #354 at all. I'd appreciate a re-review for sanity.

tonyd256:
I think all those doc examples need to use `try?` instead of `try` but otherwise :+1:

gfontenot:
No, the examples are using a `do`/`catch` block, so I believe `try` is correct here. The `Optional` is coming from the `as?` cast..

PR #341, opened 2016-04-02
Generalize the type for `sequence` taking a dict

gfontenot:
There wasn't actually any real reason to limit this to working on
`[String: Decoded<T>]` values. Instead, we should be able to use this
function with any key type.

Another change that came about because of having trouble writing documentation.
Fun stuff.


gfontenot:
Note that this is a non-breaking change 🎉

tonyd256:
:+1: love it

PR #344, opened 2016-04-02
Add inline documentation for DecodeError

gfontenot:


edwardloveall:
Looks fine (except for the 3 `/`). Only question is why inline docs and not docs at the top like elsewhere in the repo?

edwardloveall:
I guess inline isn't the right word because this is all inline, but line-by-line as opposed to grouped before the class/function/etc.

gfontenot:
`///` is used for single line doc comments, where `/** */` is used for multi-line doc comments. Given that I didn't need to add much more info about this type I felt that was adequate. I guess the argument could be made that this type is fairly self documenting, and so doesn't need this at all?

edwardloveall:
Cool, yeah seems good to me. I'm still only have little swift experience and have _no_ experience writing libraries. This is totally reasonable. :+1:

tonyd256:
:+1: unless you want periods after these

PR #346, opened 2016-04-02
Fix documentation style for FailureCoalescing

gfontenot:
We want to be consistent with our line length and whatnot.


tonyd256:
:+1:

PR #347, opened 2016-04-02
Update documentation for Decoded/Functor

gfontenot:


Argo/Types/Decoded/Functor.swift
```diff
@@ -14,6 +16,18 @@ public func <^> <T, U>(@noescape f: T -> U, x: Decoded<T>) -> Decoded<U> {
...
+  /**
+    Conditionally map a function over `self`.
+
+    - If `self` is `.Failure`, the function will not be evaluated and this will
+      return `.Failure`.
+    - If `self` is `.Success`, the function will be applied to the unwrapped
+      value.
+
+    - parameter f: A transformation function from type `T` to type `U`
+
```
tonyd256: should there be a space here? I think you've been leaving those out

PR #353, opened 2016-04-03
Add inline documentation for Decoded operators

gfontenot:


Argo/Operators/DecodeDecoded.swift
```diff
@@ -1,43 +1,150 @@
-// MARK: Values
+/**
+  Attempt to decode a value at the specified key into the requested type.
 
-// Pull value from JSON
+  This operator is used to decode a mandatory value from the JSON. If the
+  decoding fails for any reason, this will result in a `Failure` being
```
tonyd256: `.Failure`?

Argo/Operators/DecodeDecoded.swift
```diff
@@ -1,43 +1,150 @@
...
 public func <||? <A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<[A]?> {
   return .optional(json <|| [key])
 }
 
-// Pull embedded array from JSON
+/**
+  Attempt to decode an array of values at the specified key path into the
+  requested type.
+
+  This operator is used to decode a mandatory array of value from the JSON. If
```
tonyd256: values?

tonyd256:
This is good but audit is for backtick use with JSON and for the period before `.Failure` and `.Success`

Argo/Operators/DecodeDecoded.swift
```diff
@@ -1,43 +1,150 @@
...
 
-// Pull value from JSON
+  This operator is used to decode a mandatory value from the JSON. If the
+  decoding fails for any reason, this will result in a `Failure` being
+  returned.
+
+  - parameter json: The `JSON` object containing the key
+  - parameter key: The key for the object to decode
+
+  - returns: A `Decoded` value representing the success or failure of the
```
klaaspieter: In #354 @tonyd256 said a new line should be added here. I'm just saying it here, because he didn't and I'm guessing it's some convention you both overlooked.

Argo/Operators/DecodeDecoded.swift
```diff
@@ -1,43 +1,150 @@
...
 public func <|? <A where A: Decodable, A == A.DecodedType>(json: JSON, key: String) -> Decoded<A?> {
   return .optional(json <| [key])
 }
 
-// Pull embedded value from JSON
+/**
+  Attempt to decode a value at the specified key path into the requested type.
+
+  This operator is used to decode a mandatory value from the JSON. If the
+  decoding fails for any reason, this will result in a `Failure` being
```
klaaspieter: `.Failure`?

Argo/Operators/DecodeDecoded.swift
```diff
@@ -1,43 +1,150 @@
...
   return json <| [key]
 }
 
-// Pull optional value from JSON
+/**
+  Attempt to decode an optional value at the specified key into the requested
+  type.
+
+  This operator is used to decode an optional value from the JSON. If the key
+  isn't present in the JSON, this will still return `Success`. However, if the
```
klaaspieter: `.Success`?

Argo/Operators/DecodeDecoded.swift
```diff
@@ -1,43 +1,150 @@
...
-// Pull embedded value from JSON
+/**
+  Attempt to decode a value at the specified key path into the requested type.
+
+  This operator is used to decode a mandatory value from the JSON. If the
+  decoding fails for any reason, this will result in a `Failure` being
+  returned.
+
+  - parameter json: The `JSON` object containing the key
+  - parameter keys: The key path for the object to decode, represented by an
```
klaaspieter: Perhaps add an example to show what a key path should look like? I don't think immediately intuitive, especially considering the normal meaning of a key path in the APIs is a dot separated string.

klaaspieter:
I noticed the same as @tonyd256 in regards to `Success` & `Failure` instead of `.Success` and `.Failure`. The first form was used so consistently that I'm unsure at this point wether it was intentional or not.

Argo/Operators/DecodeDecoded.swift
gfontenot: I think this is probably covered by larger documentation? I have a hard time deciding what should go in here and what shouldn't. I think for now I'll keep it as is, we can add it if needed later on.

PR #356, opened 2016-04-04
Documentation for common gotchas

mbrandonw:
Inspired by a discussion on https://github.com/thoughtbot/Curry/pull/23 I have written up documentation on some common gotchas with using Argo. I put these into a separate markdown document that is linked under the "Basic usage" section.

Lemme know if this is on the right track of what y'all would expect from such a document. up for any kind of revisions to correct tone or meaning. There are probably more gotchas that could be added later. I mostly added things I've run into and scanned issues for some common mistakes.

NB: Please ignore all the commits... I used github's web interface to make these edits.

### DIFF
--- a/Documentation/Common-Gotchas.md
+++ b/Documentation/Common-Gotchas.md
@@ -9,7 +9,7 @@ to address some of those situations and set you on the right track!
 
 It is very important to fully curry your model’s `init` in order to use Argo.
 Because of this, `Argo` recommends using
-[`Curry`](https://github.com/thoughtbot/Curry) whichs allows you to change your
+[`Curry`](https://github.com/thoughtbot/Curry) which allows you to change your
 standard `init` functions into fully curried functions. Unfortunately, Swift
 can only compile `curry` for up to 20 arguments in a reasonable amount of time.
 That may sound like a lot, but one can still run up against this limitation
--- a/Documentation/Common-Gotchas.md
+++ b/Documentation/Common-Gotchas.md
@@ -1,0 +1,232 @@
+# Common Gotchas
+
+There are some common errors that arise when trying to use Argo. Many are due to
+the fact that Argo is so fundamentally different from how you might be used to
+decoding JSON, and others are due to limitations of Swift. This document aims
+to address some of those situations and set you on the right track!
+
+## `curry` limitations
+
+It is very important to fully curry your model’s `init` in order to use Argo.
+Because of this, `Argo` recommends using
+[`Curry`](https://github.com/thoughtbot/Curry) whichs allows you to change your
+standard `init` functions into fully curried functions. Unfortunately, Swift
+can only compile `curry` for up to 20 arguments in a reasonable amount of time.
+That may sound like a lot, but one can still run up against this limitation
+quite quickly, especially when you are not in control of the JSON API you are
+interfacing with.
+
+The solution is to try to find a way to normalize your data into nested models,
+thus reducing the number of arguments needed for `init`. For simplicity, assume
+for a moment that `curry` was only defined for functions of 5 arguments or less,
+and consider the following model:
+
+```swift
+struct User: Decodable {
+  let id: Int
+  let name: String
+  let bio: String
+  let smallAvatar: String
+  let mediumAvatar: String
+  let largeAvatar: String
+
+  static func decode(j: JSON) -> Decoded<User> {
+    return curry(User.init)
+      <^> j <| "id"
+      <*> j <| "name"
+      <*> j <| "bio"
+      <*> j <| "small_avatar"
+      <*> j <| "medium_avatar"
+      <*> j <| "large_avatar"
+  }
+}
+```
+
+With our assumption on `curry`, we would get a Swift compiler error when doing
+`curry(User.init)` because `User.init` takes 6 arguments. Even though the API
+has decided to send us back separate fields for each avatar size, there is a
+very clear normalization we can do in our model:
+
+```swift
+struct User {
+  let id: Int
+  let name: String
+  let bio: String
+  let avatar: Avatar
+
+  struct Avatar {
+    let small: String
+    let medium: String
+    let large: String
+  }
+}
+```
+
+Now `User.init` has 4 arguments and `User.Avatar.init` has 3 arguments, so we
+are within the assumed limitations of `curry`. But, it is also less clear how we
+should write `User.decode`. How can we extract the 3 avatar fields from `JSON`
+to build `User.Avatar` first, and then use that to build `User`? The key insight
+is to see that `j <| "key"` returns a `Decoded<T>` value, representing one step
+of decoding. So, we can decode a `User.Avatar` first using `j`, and then just
+plug that directly into `curry(User.init)`:
+
+```swift
+extension User.Avatar: Decodable {
+  static func decode(j: JSON) -> Decoded<User.Avatar> {
+    return curry(User.Avatar.init)
+      <^> j <| "small_avatar"
+      <*> j <| "medium_avatar"
+      <*> j <| "large_avatar"
+  }
+}
+
+extension User: Decodable {
+  static func decode(j: JSON) -> Decoded<User> {
+    return curry(User.init)
+      <^> j <| "id"
+      <*> j <| "name"
+      <*> j <| "bio"
+      <*> User.Avatar.decode(j)
+  }
+}
+```
+
+We have now successfully gotten around `curry`’s limitations, but also produced
+a better, normalized model! We also retain all of the great error handling that
+Argo provides, so if there is a decoding error in `User.Avatar` it will
+propogate through to a decoding error in `User`.
+
+## Compilation errors
+
+Swift’s compile errors can be mystifying if you are not intimately familiar with
+the internals of Argo. Here we describe some of the most common errors and how
+to fix them.
+
+### Complex expressions
+
+Considering that the implementation of `decode` on a model consists of fully
+currying a method and composing it with many small pieces, each of which may be
+the composition of pieces, it is not surprising that Swift sometimes has
+difficulty compiling the expression in a reasonable amount of time. This
+manifests itself with the following error:
+
+> Expression was too complex to be solved in reasonable time; consider breaking
+up the expression into distinct sub-expressions.
+
+Fortunately this is easy to solve. A first step is to simply store
+`curry(Model.init)` in a variable instead of using it directly:
+
+```swift
+static func decode(j: JSON) -> Decoded<Model> {
+  let create = Curry(Model.init)
+  return create
+    <^> j <| "key1"
+    <*> j <| "key2"
+    ...
+```
+
+Many times this can fix the problem. If you still have complex expression
+issues, the next step is to pick a midpoint in your decoding chain and set an
+intermediate variable:
+
+```swift
+static func decode(j: JSON) -> Decoded<Model> {
+  let create = Curry(Model.init)
+  let tmp = create
+    <^> j <| "key_a_1"
+    <*> j <| "key_a_2"
+    ...
+  return tmp
+    <*> j <| "key_b_1"
+    <*> j <| "key_b_2"
+    ...
+```
+
+This should almost surely fix all of your complex expression issues, but if not,
+be prepared to try one more intermediate variable. Take careful note that in
+defining `tmp` we used the map operator `create <^>` but in the return statement
+we used the applicative operator `tmp <*>`. The map operator is _only_ used with
+the fully curried `init`.
+
+### Incorrectly decoded fields
+
+When you have not correctly decoded all fields of your model in the `decode`
+function you will see an error of the form:
+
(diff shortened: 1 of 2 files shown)


tonyd256:
@mbrandonw Thanks man this is awesome!!!!

One quick comment, would you mind creating new lines at 80 characters per line? This would make it easier for us to leave you more focused comments.

mbrandonw:
@tonyd256 for sure, doing that and fixing a few mistakes. brb

mbrandonw:
@tonyd256  Ok done, and squashed commits.

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
+# Common Gotchas
+
+There are some common errors that arise when trying to use Argo. Many are due to
+ the fact that Argo is so fundamentally different from how you might be used to
```
tonyd256: looks like this line and the ones below are all indented by a space

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+ decoding JSON, and others are due to limitations of Swift. This document aims
+ to address some of those situations and set you on the right track!
+
+## `curry` limitations
+
+It is very important to fully curry your model’s `init` in order to use Argo.
+Because of this, `Argo` has a dependency on
```
tonyd256: Argo isn't actually dependent on Curry. Perhaps say that it is much easier to use with the curry library

mbrandonw: ah ha, good point!

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Complex expressions
+
+Considering that the implementation of `decode` on a model consists of fully
+currying a method and composing it with many small pieces, each of which may be
+the composition of pieces, it is not surprising that Swift sometimes has
+difficulty compiling the expression in a reasonable amount of time. This
+manifests itself with the following error:
+
+> Expression was too complex to be solved in reasonable time; consider breaking
```
tonyd256: what's up with the font on this thing?

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+The latter happens when you are returning `curry(Model.init)` directly, and the
+former happens when you have stored `curry(Model.init)` in a variable instead of
+returning directly. See the discussion on complex expressions for more
+information on why one would do that.
+
+There are a few ways in which you may not have correctly decoded your model:
+
+1. You are missing a `j <| "key"` for a field in your model.
+2. You incorrectly decoded the type of a field, e.g. you used
+`j <| "optional_key"` when you meant `j <|? "optional_key"`.
```
tonyd256: This might need to be indented

tonyd256: nope nvm, looks good when rendered

tonyd256:
@mbrandonw this is looking great. We should leave this open for a bit to get feedback from the rest of @thoughtbot/ios team and from @thoughtbot/copy-editing

Documentation/Common-Gotchas.md
mbrandonw: fixed

Documentation/Common-Gotchas.md
mbrandonw: changed to:

> Because of this, `Argo` recommends using...

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+There are some common errors that arise when trying to use Argo. Many are due to
+the fact that Argo is so fundamentally different from how you might be used to
+decoding JSON, and others are due to limitations of Swift. This document aims
+to address some of those situations and set you on the right track!
+
+## `curry` limitations
+
+It is very important to fully curry your model’s `init` in order to use Argo.
+Because of this, `Argo` recommends using
+[`Curry`](https://github.com/thoughtbot/Curry) whichs allows you to change your
```
jnutting: `whichs` typo

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+
+There are some common errors that arise when trying to use Argo. Many are due to
+the fact that Argo is so fundamentally different from how you might be used to
+decoding JSON, and others are due to limitations of Swift. This document aims
+to address some of those situations and set you on the right track!
+
+## `curry` limitations
+
+It is very important to fully curry your model’s `init` in order to use Argo.
+Because of this, `Argo` recommends using
```
gfontenot: What do you think about changing this to say "we recommend"? Feels kinda weird to take advice from a library.

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+[`Curry`](https://github.com/thoughtbot/Curry) which allows you to change your
+standard `init` functions into fully curried functions. Unfortunately, Swift
+can only compile `curry` for up to 20 arguments in a reasonable amount of time.
```
gfontenot: how about "for around 20 arguments"? I'd like to avoid needing to update this doc if we make incremental improvements to `curry`.

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+      <*> j <| "name"
+      <*> j <| "bio"
+      <*> j <| "small_avatar"
+      <*> j <| "medium_avatar"
+      <*> j <| "large_avatar"
+  }
+}
+```
+
+With our assumption on `curry`, we would get a Swift compiler error when doing
```
gfontenot: It might be nice to have an example of what this error looks like, because it sure doesn't look like you'd expect.

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+`curry(User.init)` because `User.init` takes 6 arguments. Even though the API
+has decided to send us back separate fields for each avatar size, there is a
+very clear normalization we can do in our model:
```
gfontenot: might avoid "very clear" here to avoid alienating people that don't find it clear at all.

gfontenot: In fact, it might be best to be more verbose and say "Even though the API has decided to send back separate fields for each avatar size, we can normalize this response by introducing a new `Avatar` type:"

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+
+```swift
+struct User {
+  let id: Int
+  let name: String
+  let bio: String
+  let avatar: Avatar
+
+  struct Avatar {
```
gfontenot: I might avoid nesting this to avoid potentially confusing people into thinking it _has_ to be nested.

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+  }
+}
+```
+
+We have now successfully gotten around `curry`’s limitations, but also produced
+a better, normalized model! We also retain all of the great error handling that
+Argo provides, so if there is a decoding error in `User.Avatar` it will
+propogate through to a decoding error in `User`.
+
+## Compilation errors
```
gfontenot: I know we talked about this being in one file, but maybe we _should_ break this into two files under a "Common Gotchas" section?

Documentation/Common-Gotchas.md
gfontenot: it's a quote

gfontenot:
All of this is really great. Made some comments, going to try to get more feedback from grammar people.

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+    return curry(User.init)
+      <^> j <| "id"
+      <*> j <| "name"
+      <*> j <| "bio"
+      <*> User.Avatar.decode(j)
+  }
+}
+```
+
+We have now successfully gotten around `curry`’s limitations, but also produced
```
mike-burns: I'd go with "and" instead of "but" here, since you're joining two clauses stating positive things.

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+Considering that the implementation of `decode` on a model consists of fully
+currying a method and composing it with many small pieces, each of which may be
+the composition of pieces, it is not surprising that Swift sometimes has
+difficulty compiling the expression in a reasonable amount of time. This
+manifests itself with the following error:
+
+> Expression was too complex to be solved in reasonable time; consider breaking
+up the expression into distinct sub-expressions.
+
+Fortunately this is easy to solve. A first step is to simply store
```
mike-burns: We could drop the word "simply" here: it would make the sentence shorter, and also alienate fewer people.

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+    <*> j <| "key_b_1"
+    <*> j <| "key_b_2"
+    ...
+```
+
+This should almost surely fix all of your complex expression issues, but if not,
+be prepared to try one more intermediate variable. Take careful note that in
+defining `tmp` we used the map operator `create <^>` but in the return statement
+we used the applicative operator `tmp <*>`. The map operator is _only_ used with
+the fully curried `init`.
```
mike-burns: Is it worth it to mention the types here?

Documentation/Common-Gotchas.md
```diff
@@ -0,0 +1,232 @@
...
+'_ -> _ -> C'
+
+The latter happens when you are returning `curry(Model.init)` directly, and the
+former happens when you have stored `curry(Model.init)` in a variable instead of
+returning directly. See the discussion on complex expressions for more
+information on why one would do that.
+
+There are a few ways in which you may not have correctly decoded your model:
+
+1. You are missing a `j <| "key"` for a field in your model.
```
mike-burns: Should this be a bulleted list (`*` or `-`) instead of the enumerated list (`1.`, `2.`)? I believe this is listing two independent options and not a sequence of steps.

mike-burns:
This is a fantastic write-up.

Documentation/Common-Gotchas.md
geoffharcourt: "Our recommended best practice it so use..."?

mbrandonw:
thanks for the notes! i'll be looking at this soon and take a pass at it

Documentation/Common-Gotchas.md
tonyd256: It appears that there is another solution here. One could also give the type inference of the compiler a break and type all the expressions. Here is an example from the `TestModel` class in our tests.

```swift
extension TestModel: Decodable {
  static func decode(_ json: JSON) -> Decoded<TestModel> {
    return curry(self.init)
      <^> json <| "numerics"                                    as Decoded<TestModelNumerics>
      <*> json <| ["user_opt", "name"]                          as Decoded<String>
      <*> json <| "bool"                                        as Decoded<Bool>
      <*> json <|| "string_array"                               as Decoded<[String]>
      <*> json <||? "string_array_opt"                          as Decoded<[String]?>
      <*> json <|| ["embedded", "string_array"]                 as Decoded<[String]>
      <*> json <||? ["embedded", "string_array_opt"]            as Decoded<[String]?>
      <*> json <|? "user_opt"                                   as Decoded<User?>
      <*> (json <| "dict" >>- { [String: String].decode($0) })  as Decoded<[String: String]>
      <*> json <| "numerics"                                    as Decoded<TestModelNumerics>
      <*> json <| ["user_opt", "name"]                          as Decoded<String>
      <*> json <| "bool"                                        as Decoded<Bool>
      <*> json <|| "string_array"                               as Decoded<[String]>
      <*> json <||? "string_array_opt"                          as Decoded<[String]?>
      <*> json <|| ["embedded", "string_array"]                 as Decoded<[String]>
      <*> json <||? ["embedded", "string_array_opt"]            as Decoded<[String]?>
      <*> json <|? "user_opt"                                   as Decoded<User?>
      <*> (json <| "dict" >>- { [String: String].decode($0) })  as Decoded<[String: String]>
      <*> json <| "numerics"                                    as Decoded<TestModelNumerics>
      <*> json <| ["user_opt", "name"]                          as Decoded<String>
      <*> json <| "bool"                                        as Decoded<Bool>
  }
}
```

This way is arguably uglier but still works without intermediate variables. It can work up to and probably past the limit of curry.

gfontenot:
@mbrandonw Do you have time to finish this up? I'd be happy to take it over and apply fixes for the comments if you don't.

mbrandonw:
> @mbrandonw Do you have time to finish this up? I'd be happy to take it over and apply fixes for the comments if you don't.

so sorry for being absent! if you could help out it would be much appreciated. i'm having a hard time finding spare time at this moment :/

gfontenot:
@mbrandonw No worries! I'm happy to take it over.

gfontenot:
Oh hey the revised version of this got merged so we can close it. Huzzah! 🎉

PR #359, opened 2016-04-06
Allow boolean values to be represented by numbers

### LINKED ISSUE #358: Argo 3 breaks implicit Number -> Bool conversion
With the addition of an actual `JSON.Bool` case, it seems that the implicit `.Number` conversion to `Bool` has been broken. Was this intentional? It's unfortunately common for backend software to return boolean values as `1` or `0`. Previously Argo was able to automatically convert these to `Bool`. Now that's no longer possible. I believe the previous behavior could be returned by adding the previous behavior back to the `Bool` `Decodable` extension:
```swift
extension Bool: Decodable {
  /**
    Decode `JSON` into `Decoded<Bool>`.

    Succeeds if the value is a boolean, otherwise it returns a type mismatch.

    - parameter j: The `JSON` value to decode

    - returns: A decoded `Bool` value
  */
  public static func decode(j: JSON) -> Decoded<Bool> {
    switch j {
    case let .Bool(n): return pure(n)
    case let .Number(n as Bool): return pure(n)
    default: return .typeMismatch("Bool", actual: j)
    }
  }
}
```
Would that be an acceptable change?

gfontenot:
Servers commonly represent boolean values as numbers even though this is
clearly wrong and they should be sending `true`/`false`.

Anyway, we can make it easier to deal with this by allowing numbers to
represent boolean values as well. This essentially fixes compatibility
with earlier versions of Argo, which did this anyway.

Fixes #358


tonyd256:
:+1: do we have a test that can demo this? and possibly a number that fails (if that exists)

jshier:
Test should be straightforward, just parse this JSON, both as `true`:
```json
{
  "realBool": true,
  "numberBool": 1
}
```
As for failure, I can't really think of one for the number case, as any number this isn't 0 (`false`) will be cast to `Bool` as `true`. Aside from overflows, which can't be parsed into a number in the first place.

gfontenot:
Added a success test and fixed the build by writing code that actually compiles (which seemed useful)

PR #361, opened 2016-04-07
Add an issue template to speed up issue resolution

tonyd256:
Github recently added the capability to add a file to use as an issue template.
I've used these before on other repos and they are very useful. We get a decent
amount of issue requests in Argo and sometimes they require us to ask for more
information.  This will hopefully help get all the information we need to
resolve most issues.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,35 @@
+_[Description of the issue you are experiencing]_
+
+### Error Messages
+
+_List all the error messages or warnings you are seeing, if any._
+
+### Sample JSON
+
+```
+{
+  "key": "value"
+}
+```
+
+### Models
+
+_All models involved, including their `Decodable` implementations._
+
+```
+struct User {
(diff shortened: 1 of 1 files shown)


tonyd256:
We can also hide this file in a `.github` folder if we don't want it to be visible in the root. I think it's fine here though. https://github.com/thoughtbot/Argo/pull/361

tonyd256:
We can also hide this file in a .github folder if we don't want it to be visible in the root. I think it's fine here though. https://help.github.com/articles/creating-an-issue-template-for-your-repository/#adding-an-issue-template

dazmuda:
### Description

### Argo Version
### Dependency Manager
### Error Messages

### Sample JSON
### Models

tonyd256:
That makes sense to me

dazmuda:
This looks good to me.

gfontenot:
Dig it. I don't mind it in the root of the repo right now, although I _am_ intrigued by the `.github` folder

PR #373, opened 2016-05-13
Turn on whole module optimizations

tonyd256:
After turning this optimization level on, I observed an average decoding
time of 1.611s from 2.523s and an average parsing time of 1.724s from
2.384s.


tonyd256:
After reading through this [optimization guide](https://github.com/apple/swift/blob/master/docs/OptimizationTips.rst) I found that whole module optimization was the only thing missing from Argo.

gfontenot:
👍

PR #376, opened 2016-06-14
WIP: Update Argo for Swift 3

tonyd256:
This is a dumb update that simply went through the motions to get Argo
compiling in Xcode 8 beta 1. This should open up discussion on what
changes we should make for Swift 3.

### DIFF
--- a/Argo/Extensions/Dictionary.swift
+++ b/Argo/Extensions/Dictionary.swift
@@ -9,7 +9,7 @@ func + <T, U>(lhs: [T: U], rhs: [T: U]) -> [T: U] {
 }
 
 extension Dictionary {
-  func map<T>(@noescape f: Value -> T) -> [Key: T] {
+  func map<T>(_ f: @noescape (Value) -> T) -> [Key: T] {
     var accum = Dictionary<Key, T>(minimumCapacity: self.count)
 
     for (key, value) in self {
@@ -20,6 +20,6 @@ extension Dictionary {
   }
 }
 
-func <^> <T, U, V>(@noescape f: T -> U, x: [V: T]) -> [V: U] {
+func <^> <T, U, V>(f: @noescape (T) -> U, x: [V: T]) -> [V: U] {
   return x.map(f)
 }
--- a/Argo/Functions/curry.swift
+++ b/Argo/Functions/curry.swift
@@ -1,3 +1,0 @@
-func curry<T, U, V>(f : (T, U) -> V) -> T -> U -> V {
-  return { x in { y in f(x, y) } }
-}
--- a/Argo/Functions/flatReduce.swift
+++ b/Argo/Functions/flatReduce.swift
@@ -17,7 +17,7 @@
   - returns: The result of iterating the combinator over every element of the
              sequence and flattening the result
 */
-public func flatReduce<S: SequenceType, U>(sequence: S, initial: U, @noescape combine: (U, S.Generator.Element) -> Decoded<U>) -> Decoded<U> {
+public func flatReduce<S: Sequence, U>(_ sequence: S, initial: U, combine: @noescape (U, S.Iterator.Element) -> Decoded<U>) -> Decoded<U> {
   return sequence.reduce(pure(initial)) { accum, x in
     accum >>- { combine($0, x) }
   }
--- a/Argo/Types/Decoded/Monad.swift
+++ b/Argo/Types/Decoded/Monad.swift
@@ -11,7 +11,7 @@
 
   - returns: A value of type `Decoded<U>`
 */
-public func >>- <T, U>(x: Decoded<T>, @noescape f: T -> Decoded<U>) -> Decoded<U> {
+public func >>- <T, U>(x: Decoded<T>, f: @noescape (T) -> Decoded<U>) -> Decoded<U> {
   return x.flatMap(f)
 }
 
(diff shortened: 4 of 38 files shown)


Argo/Extensions/Dictionary.swift
```diff
@@ -20,6 +20,6 @@ extension Dictionary {
   }
 }
 
-func <^> <T, U, V>(@noescape f: T -> U, x: [V: T]) -> [V: U] {
+func <^> <T, U, V>( f: @noescape (T) -> U, x: [V: T]) -> [V: U] {
```
gfontenot: Missing a `_` here?

Argo/Functions/flatReduce.swift
```diff
@@ -17,7 +17,7 @@
   - returns: The result of iterating the combinator over every element of the
              sequence and flattening the result
 */
-public func flatReduce<S: SequenceType, U>(sequence: S, initial: U, @noescape combine: (U, S.Generator.Element) -> Decoded<U>) -> Decoded<U> {
+public func flatReduce<S: Sequence, U>(sequence: S, initial: U, combine: @noescape (U, S.Iterator.Element) -> Decoded<U>) -> Decoded<U> {
```
gfontenot: Do we need to do `_ sequence` here?

Argo/Extensions/Dictionary.swift
tonyd256: ah, no we don't need them on operators ... but the automatic conversion tool left that extra space in there so I need to fix that

Argo/Types/Decoded/Monad.swift
```diff
@@ -28,7 +28,7 @@ public func >>- <T, U>(x: Decoded<T>, @noescape f: T -> Decoded<U>) -> Decoded<U
 
   - returns: A value of type `Decoded<U>`
 */
-public func -<< <T, U>(@noescape f: T -> Decoded<U>, x: Decoded<T>) -> Decoded<U> {
+public func -<< <T, U>( f: @noescape (T) -> Decoded<U>, x: Decoded<T>) -> Decoded<U> {
```
gfontenot: There's a weird space before `f` here

Argo/Functions/curry.swift
```diff
@@ -1,3 +1,3 @@
-func curry<T, U, V>(f : (T, U) -> V) -> T -> U -> V {
+func curry<T, U, V>(f : (T, U) -> V) -> (T) -> (U) -> V {
```
tonyd256: just realized we're not using this anymore ... will delete

Argo/Types/JSON.swift
```diff
@@ -45,19 +45,20 @@ public extension JSON {
 }
 
 extension JSON: Decodable {
+  public typealias DecodedType = JSON
```
gfontenot: Why is this needed all of a sudden?

tonyd256: good catch .. it's not .. was messing around with a bug and forgot to remove

ArgoTests/JSON/JSONFileReader.swift
```diff
@@ -1,13 +1,15 @@
 import Foundation
 
 func JSONFromFile(file: String) -> AnyObject? {
-  return NSBundle(forClass: JSONFileReader.self).pathForResource(file, ofType: "json")
-    .flatMap { NSData(contentsOfFile: $0) }
+  Bundle().pathForResource("user_without_key", ofType: "json")
```
gfontenot: What's going on here?

ArgoTests/JSON/JSONFileReader.swift
```diff
@@ -1,13 +1,15 @@
...
+  return Bundle(for: JSONFileReader.self).pathForResource(file, ofType: "json")
+    .flatMap { URL(string: "file://\($0)") }
```
gfontenot: This is unfortunate

Argo/Functions/flatReduce.swift
tonyd256: maybe ... I intentionally left some function without `_`s to have a discussion about if it should since that's the way swift seems to be headed

gfontenot:
Few comments, nothing major. I think this looks good in general. We _do_ need to update our enum naming to match Swift 3.0 IMO. It's going to take _forever_ for me to get used to lowercased enum cases but it's worth being consistent.

ArgoTests/JSON/JSONFileReader.swift
tonyd256: ya looks like we lost that `NSData(contentsOfFile: String)` constructor ... and if we use the URL constructor then it needs a `file://` in front

ArgoTests/JSON/JSONFileReader.swift
tonyd256: yup ... left in there by accident

Argo/Functions/flatReduce.swift
gfontenot: I think we should maintain the existing API here. I'd vote for ignoring the external label on `sequence`.

ArgoTests/JSON/JSONFileReader.swift
```diff
@@ -1,13 +1,14 @@
 import Foundation
 
 func JSONFromFile(file: String) -> AnyObject? {
-  return NSBundle(forClass: JSONFileReader.self).pathForResource(file, ofType: "json")
-    .flatMap { NSData(contentsOfFile: $0) }
+  return Bundle(for: JSONFileReader.self).pathForResource(file, ofType: "json")
+    .flatMap { URL(string: "file://\($0)") }
```
gfontenot: We can use `URL(fileURLWithPath:)` here (pretty sure that's what it's called)

gfontenot:
I'm OK with these changes. Was there anything specific you weren't sure about? I'd like to get this merged into `master` (along with a disclaimer about version compatibility) ASAP so that we can move forward with other Swift 3.0 changes that need to take place.

PR #379, opened 2016-06-18
Set SWIFT_VERSION = 2.3 (for Xcode 8)

sharplet:
Depends on https://github.com/thoughtbot/Curry/pull/27. See that PR for rationale.

To do:

- [ ] Update to a tagged release of Curry


sharplet:
It could also be worth having a build matrix of both Xcode 7.3 and Xcode 8. Thoughts?

sharplet:
Because master has already diverged for Swift 3, I'm not sure what the best course of action is regarding this branch. It might make sense to create a `swift-2` branch (or repurpose this branch) and have it be long-lived.

sharplet:
Ok, so here's the failure on CI:

```
xcodebuild: error: Unable to find a destination matching the provided destination specifier:
		{ platform:tvOS Simulator, name:Apple TV 1080p }

	The requested device could not be found because no available devices matched the request.
```

I'm not sure why Circle can't find the Apple TV simulator. All the tests pass on my machine, under both Xcode 7 & 8.

gfontenot:
That's a problem with Circle. I alerted them about it like a month ago, still no word on a fix.

Cartfile.private
```diff
@@ -1,2 +1,2 @@
-github "thoughtbot/Curry"
+github "thoughtbot/Curry" ~> 2.0
```
ikesyo: Xcode 8 support is only in Curry v2.3.2, so this should be `~> 2.3.2`.

ikesyo:
I want a new release with this (for Xcode 8 support in Tentacle and Carthage), which supports both Swift2.2/Xcode 7.3 and Swift 2.3/Xcode 8. Is there any chance to get this merged?

gfontenot:
So, to be clear, what we need here is this commit applied on top of the 3.0.1 release, and that should then be released as 3.0.2. Is that correct? /cc @sharplet @ikesyo

ikesyo:
> So, to be clear, what we need here is this commit applied on top of the 3.0.1 release, and that should then be released as 3.0.2. Is that correct?

That seems correct. Applying this onto 13db214 instead of the `v3.0.1` tag (be127e5) might be better.

gfontenot:
Cool. I'll see if I can get this out today.

ikesyo:
@gfontenot Thanks for the new release! However, I've realized that I'd like to get some additional patches for better Xcode 8 support. It would be great if you prepare `swift2.3` or something similar branch and then I will submit a PR onto that branch. Of should I target that onto this existing `swift-2.3` branch?

I'd created a branch for that and you can check the diff here:
- https://github.com/ikesyo/Argo/tree/swift2.3-xcode8-support-improvements
- https://github.com/thoughtbot/Argo/compare/v3.0.2...ikesyo:swift2.3-xcode8-support-improvements

gfontenot:
@ikesyo I just created an `xcode-8` branch, feel free to open a PR against that and I'll try to get it in asap.

ikesyo:
Really appreciate it!

PR #381, opened 2016-06-25
WIP: Improve speed with multithreading

tonyd256:
By using an `OperationQueue` and pre-allocating memory in our arrays we are able to significantly increase the speed of JSON parsing and decoding.

Currently on master, parsing averages 0.973s and decoding averages 0.79s with a combined average of 1.763s. With the changes on this branch we were able to achieve a parsing average of 0.393s and a decoding average of 0.294s with a combined average of 0.687s. This is a 61% increase in speed.

I haven't noticed a significant difference between 10 or 100 threads (by setting `divider`) but 100 did seem faster. I'm checking if an array is over 100 before deciding to use threading but that means that if there are 101 items, we'll try to use 100 threads to decode them which probably isn't the most efficient.
I think there is still some work to be done around when to use threading, how many threads to use, and how many items should be processed in each thread. However, this is pretty cool how it is now.

### DIFF
--- a/Argo/Types/Task.swift
+++ b/Argo/Types/Task.swift
@@ -1,0 +1,54 @@
+import Foundation
+
+private let queue = OperationQueue()
+
+final class Task<U, T> {
+  var array: ArraySlice<U>
+  var result: [T]
+  let transform: (U) -> T
+
+  init(_ array: ArraySlice<U>, _ transform: (U) -> T) {
+    self.array = array
+    result = []
+    result.reserveCapacity(array.count)
+    self.transform = transform
+  }
+
+  func perform() {
+    result = array.map(transform)
+  }
+}
+
+private let divider = 100
+
+func divideAndConquer<U, T>(input: [U], transform: (U) -> T) -> [T] {
+  let count = Int(ceil(Double(input.count) / Double(divider)))
+
+  var final: [T] = []
+  final.reserveCapacity(input.count)
+
+  var tasks: [Task<U, T>] = []
+  tasks.reserveCapacity(divider)
+
+  for i in 0..<divider {
+    let startIndex = i*count
+    let endIndex = min(i*count+count, input.endIndex) - 1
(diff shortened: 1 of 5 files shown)


Argo/Types/Task.swift
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  func perform() {
+    result = array.map(transform)
+  }
+}
+
+private let divider = 100
+
+func divideAndConquer<U, T>(input: [U], transform: (U) -> T) -> [T] {
```
jakecraige: 👏  😍

Argo/Types/Task.swift
```diff
@@ -0,0 +1,54 @@
...
+    let task = Task<U, T>(slice, transform)
+    tasks.append(task)
+  }
+
+  queue.maxConcurrentOperationCount = divider
+
+  tasks.forEach { task in
+    queue.addOperation(task.perform)
+  }
+  queue.waitUntilAllOperationsAreFinished()
```
jakecraige: Would using `queue.addOperations(tasks.map({ $0.perform }), waitUntilFinished: true)` work here?

jakecraige: It looks like this is Swift 3 stuff, I'm not sure if that API exists in it or not

gfontenot:
To make those numbers a bit easier to see:

|  | parsing | decoding | total |
| --- | --- | --- | --- |
| before | 0.973s | 0.79s | 1.763 |
| after | 0.393s | 0.294s | 0.687s |
| improvement | 59.6% | 62% | 61% |

gfontenot:
What are we doing here? Do we still want to move forward with this?

eimantas:
I'd really like some performance improvement. Now `decode` functions take first place when profiling function body compile times...

gfontenot:
We definitely want to do some work here, but this PR is crazy out of date. I'm going to close this until we have the bandwidth to tackle this properly.

PR #382, opened 2016-06-27
Add project specific vimrc

gfontenot:
This vimrc defines our local tab preference, as well as specifies a default
scheme to be used by [xcode.vim](https://github.com/gfontenot/vim-xcode).

You can tell vim to read this local vimrc by default by ensuring that you have
the following in your main vimrc:

```
set secure  " Don't let external configs do scary shit
set exrc    " Load local vimrc if found
```

### DIFF
--- a/.vimrc
+++ b/.vimrc
@@ -1,0 +1,4 @@
+set tabstop=2
+set shiftwidth=2
+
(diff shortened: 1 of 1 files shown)


tonyd256:
👍

PR #384, opened 2016-07-01
Remove playground

gfontenot:
I think this was a great idea, but ultimately was annoying to keep up to date
and didn't serve its purpose properly. Styles changed and broke constantly
between Xcode releases, and it never _quite_ felt right. I think this kind of
stuff would probably be better served as documentation (something like literal
swift would be perfect) in the long run, but for now I think we should just
remove it to avoid confusion.


tonyd256:
👍

PR #388, opened 2016-07-11
Use Semigroup type to coalesce multiple errors

klaaspieter:
Inspired by this Haskell [Validation type]. Which is essentially the same as `Either` but with an accumulating left hand side. This applied to `Decoded` is a `Decoded` with an accumulating `DecodedError`.

This allows consumers to be warned about all the JSON errors, instead of having to fix each one by one.

[Validation type]: https://github.com/tonymorris/validation

### DIFF
--- a/Argo/Types/Decoded/Applicative.swift
+++ b/Argo/Types/Decoded/Applicative.swift
@@ -47,9 +47,10 @@ public extension Decoded {
     - returns: A value of type `Decoded<U>`
   */
   func apply<U>(_ f: Decoded<(T) -> U>) -> Decoded<U> {
-    switch f {
-    case let .success(function): return self.map(function)
-    case let .failure(error): return .failure(error)
+    switch (f, self) {
+    case let (.success(function), _): return self.map(function)
+    case let (.failure(le), .failure(re)): return .failure(le + re)
+    case let (.failure(f), _): return .failure(f)
     }
   }
 }
--- a/ArgoTests/Tests/DecodedTests.swift
+++ b/ArgoTests/Tests/DecodedTests.swift
@@ -55,7 +55,44 @@ class DecodedTests: XCTestCase {
     default: XCTFail("Unexpected Case Occurred")
     }
   }
-  
+
+  func testDecodedMultipleErrors() {
+    let user: Decoded<User> = decode([
+      "id": "1"
+    ])
+
+    let expected: [DecodeError] = [
+      .typeMismatch(expected: "Int", actual: "String(1)"),
+      .missingKey("name")
+    ]
+
+    switch user {
+    case let .failure(.multiple(errors)): XCTAssert(errors == expected)
+    default: XCTFail("Unexpected Case Occurred")
+    }
+  }
+
+  func testDecodedMultipleErrorsWithOptionalKeyTypeMismatch() {
+    let user: Decoded<User> = decode([
+      "id": 1,
+      "email": 1
+    ])
+
+    let expected: [DecodeError] = [
+      .missingKey("name"),
+      .typeMismatch(expected: "String", actual: String(describing: JSON.number(1)))
+    ]
+
+    switch user {
+    case let .failure(.multiple(errors)):
+      print("expected: \(expected)")
+      print("actual: \(errors)")
+
+      XCTAssert(errors == expected)
+    default: XCTFail("Unexpected Case Occurred")
+    }
+  }
+
   func testDecodedMaterializeSuccess() {
     let user: Decoded<User> = decode(json(fromFile: "user_with_email")!)
     let materialized = materialize { user.value! }
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -546,8 +546,8 @@
 				EA395DC31A52F8EB00EB607E /* array_root.json */,
 				EA395DC91A52FC1400EB607E /* root_object.json */,
 				EA6DD69E1AB384C700CA3A5B /* big_data.json */,
-				EA47BB551AFC5DAC002D2CCD /* user_with_bad_type.json */,
 				EA47BB581AFC5E65002D2CCD /* user_without_key.json */,
+				EA47BB551AFC5DAC002D2CCD /* user_with_bad_type.json */,
 			);
 			path = JSON;
 			sourceTree = "<group>";
--- a/Argo/Types/DecodeError.swift
+++ b/Argo/Types/DecodeError.swift
@@ -8,6 +8,9 @@ public enum DecodeError: Error {
 
   /// A custom error case for adding explicit failure info.
   case custom(String)
+
+  /// There were multiple errors in the JSON.
+  case multiple([DecodeError])
 }
 
 extension DecodeError: CustomStringConvertible {
@@ -16,6 +19,7 @@ extension DecodeError: CustomStringConvertible {
     case let .typeMismatch(expected, actual): return "TypeMismatch(Expected \(expected), got \(actual))"
     case let .missingKey(s): return "MissingKey(\(s))"
     case let .custom(s): return "Custom(\(s))"
+    case let .multiple(es): return "Multiple(\(es.map { $0.description }.joined(separator: ", ")))"
     }
   }
 }
@@ -29,6 +33,8 @@ extension DecodeError: Hashable {
       return string.hashValue
     case let .custom(string):
       return string.hashValue
+    case let .multiple(es):
+      return es.reduce(0) { $0 ^ $1.hashValue }
     }
   }
 }
@@ -44,7 +50,18 @@ public func == (lhs: DecodeError, rhs: DecodeError) -> Bool {
   case let (.custom(string1), .custom(string2)):
     return string1 == string2
 
+  case let (.multiple(lhs), .multiple(rhs)):
+    return lhs == rhs
+
   default:
     return false
+  }
+}
+
+public func + (lhs: DecodeError, rhs: DecodeError) -> DecodeError {
+  switch (lhs, rhs) {
+  case let (.multiple(es), e): return .multiple(es + [e])
+  case let (e, .multiple(es)): return .multiple([e] + es)
+  case let (le, re): return .multiple([le, re])
   }
 }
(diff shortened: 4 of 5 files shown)


tonyd256:
great stuff!

This does change some things about how Argo works and will probably need doc updates in many places. Also, if we're not failing fast anymore, I wonder if there are any performance implications.

klaaspieter:
@tonyd256 I removed the semigroup abstraction. 

I did a quick glance of the docs and couldn't find any obvious places where we're explicitly saying only the first error will be returned.

As for performance, I ran the performance tests and they still succeeded. Anything else I should check?

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -55,7 +55,22 @@ class DecodedTests: XCTestCase {
...
-  
+
+  func testDecodedMultipleErrors() {
+    let user: Decoded<User> = decode(json(fromFile: "user_with_bad_type_and_missing_key")!)
+
+    let expected: [DecodeError] = [
+      .typeMismatch(expected: "Int", actual: "String(1)"),
+      .missingKey("name")
+    ]
+    print(expected)
```
tonyd256: do we need to keep this?

tonyd256:
I had one small removal but otherwise this is looking good. This does feel like a major change to how Argo has worked in the past and I would like @gfontenot opinion here as well.

ArgoTests/Tests/DecodedTests.swift
gfontenot: No, let's not check this in

gfontenot:
Yeah, this feels like a major change, but I kinda like it? I'm a _little_ worried that users have to explicitly match on single vs multiple errors, but I'm not sure how else to handle that. I'm also unsure how often users would _actually_ do that.

Argo/Types/Decoded/Applicative.swift
```diff
@@ -45,9 +45,12 @@ public extension Decoded {
...
   func apply<U>(_ f: Decoded<(T) -> U>) -> Decoded<U> {
-    switch f {
-    case let .success(function): return self.map(function)
-    case let .failure(error): return .failure(error)
+    switch (f, self) {
+    case let (.success(function), _): return self.map(function)
+    case let (.failure(.multiple(es)), .failure(e)): return .failure(.multiple(es + [e]))
+    case let (.failure(e), .failure(.multiple(es))): return .failure(.multiple([e] + es))
+    case let (.failure(le), .failure(re)): return .failure(.multiple([le, re]))
+    case let (.failure(f), _): return .failure(f)
```
gfontenot: Had a thought about this? What do you think about implementing `+` for `DecodeError` and using it here instead of matching on the specific error case?

tonyd256:
I can't say I've actually used the error messaging very much but I can see this being especially useful when getting started. I definitely think it's a breaking change so should be a major version change but that will probably happen with swift 3 anyways.

klaaspieter:
@tonyd256 this is the output from the performance tests on master:

```
Test Suite 'Selected tests' started at 2016-08-10 08:40:57.115
Test Suite 'Argo-MacTests.xctest' started at 2016-08-10 08:40:57.116
Test Suite 'PerformanceTests' started at 2016-08-10 08:40:57.117
Test Case '-[Argo_MacTests.PerformanceTests testBigDataDecodesCorrectly]' started.
Test Case '-[Argo_MacTests.PerformanceTests testBigDataDecodesCorrectly]' passed (2.804 seconds).
Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' started.
/Users/klaaspieter/Developer/thoughtbot/Argo/ArgoTests/Tests/PerformanceTests.swift:10: Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' measured [Time, seconds] average: 1.452, relative standard deviation: 4.660%, values: [1.583406, 1.397201, 1.398312, 1.409109, 1.440552, 1.422274, 1.413303, 1.430184, 1.437901, 1.584596], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' passed (15.101 seconds).
Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' started.
/Users/klaaspieter/Developer/thoughtbot/Argo/ArgoTests/Tests/PerformanceTests.swift:19: Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' measured [Time, seconds] average: 0.937, relative standard deviation: 1.613%, values: [0.974236, 0.923664, 0.933166, 0.950622, 0.935832, 0.927375, 0.927761, 0.925622, 0.927496, 0.948155], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' passed (11.508 seconds).
Test Suite 'PerformanceTests' passed at 2016-08-10 08:41:26.532.
	 Executed 3 tests, with 0 failures (0 unexpected) in 29.413 (29.415) seconds
Test Suite 'Argo-MacTests.xctest' passed at 2016-08-10 08:41:26.533.
	 Executed 3 tests, with 0 failures (0 unexpected) in 29.413 (29.416) seconds
Test Suite 'Selected tests' passed at 2016-08-10 08:41:26.533.
	 Executed 3 tests, with 0 failures (0 unexpected) in 29.413 (29.418) seconds
```

This is the output for this branch:

```
Test Suite 'Selected tests' started at 2016-08-10 08:45:29.167
Test Suite 'Argo-MacTests.xctest' started at 2016-08-10 08:45:29.167
Test Suite 'PerformanceTests' started at 2016-08-10 08:45:29.168
Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' started.
/Users/klaaspieter/Developer/thoughtbot/Argo/ArgoTests/Tests/PerformanceTests.swift:10: Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' measured [Time, seconds] average: 1.414, relative standard deviation: 4.939%, values: [1.429404, 1.371464, 1.577351, 1.505955, 1.350685, 1.362656, 1.350375, 1.394410, 1.404189, 1.393458], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[Argo_MacTests.PerformanceTests testParsePerformance]' passed (14.626 seconds).
Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' started.
/Users/klaaspieter/Developer/thoughtbot/Argo/ArgoTests/Tests/PerformanceTests.swift:19: Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' measured [Time, seconds] average: 1.062, relative standard deviation: 6.006%, values: [1.070299, 0.992940, 1.024443, 1.086184, 1.038224, 0.976535, 1.004668, 1.103101, 1.179291, 1.147239], performanceMetricID:com.apple.XCTPerformanceMetric_WallClockTime, baselineName: "", baselineAverage: , maxPercentRegression: 10.000%, maxPercentRelativeStandardDeviation: 10.000%, maxRegression: 0.100, maxStandardDeviation: 0.100
Test Case '-[Argo_MacTests.PerformanceTests testDecodePerformance]' passed (12.511 seconds).
Test Case '-[Argo_MacTests.PerformanceTests testBigDataDecodesCorrectly]' started.
Test Case '-[Argo_MacTests.PerformanceTests testBigDataDecodesCorrectly]' passed (2.833 seconds).
Test Suite 'PerformanceTests' passed at 2016-08-10 08:45:59.139.
	 Executed 3 tests, with 0 failures (0 unexpected) in 29.969 (29.971) seconds
Test Suite 'Argo-MacTests.xctest' passed at 2016-08-10 08:45:59.140.
	 Executed 3 tests, with 0 failures (0 unexpected) in 29.969 (29.972) seconds
Test Suite 'Selected tests' passed at 2016-08-10 08:45:59.140.
	 Executed 3 tests, with 0 failures (0 unexpected) in 29.969 (29.973) seconds
```

klaaspieter:
This should be all your feedback @tonyd256 @gfontenot. I had to sneak in 7b70caf49f473922e1c481b7f47e1ce72b547430 to get Argo to build in Xcode 8 beta 4. I can PR that separately or commit to master directly if you prefer.

gfontenot:
If we're going to do this we need to decide now before we ship the next version. @tonyd256 @klaaspieter @thoughtbot/ios thoughts?

tonyd256:
In general I'm for this. I feel a bit nervous about this causing people's code to not work. The only case I can think of is if someone is checking for a specific error case like `.typeMismatch` or `.missingKey` and now those cases won't be true because they may be hidden within `.multiple`.

I'm also not clear if this affects the case where `<|?` is ok with `.missingKey` and not ok with `.typeMismatch`. What happens when there is a mix?

klaaspieter:
For the first I think it's good practice to avoid using `default` in a `switch` statement, because it'll error if you're not handling all possible cases. It is however a breaking change so would require a version bump, and perhaps a warning in the release notes about that specific case.

For the second one do you have a specific example in mind? The reason the `testDecodedMultipleErrors` uses the `user_with_bad_type_and_missing_key.json` with an email key was to ensure that missing optional keys aren't part of the coalesced errors. 

I have a local commit that adds an extra test to ensure the `typeMismatch` works with optional keys. I'll push it if someone can re-add me to the Argo repo. Otherwise I can do a new PR from a branch on my personal fork.

tonyd256:
@gfontenot So I think this code was actually good to go.

I made a change that I think actually handles the issue in #401  ... I'm going to remove those commits now and put them in a different branch.

tonyd256:
After digging into another issue I think I found that Aeson does not return multiple errors if there are such. Not sure if that tips the discussion.

PR #390, opened 2016-07-18
Update for Swift 3 & Xcode 8 b3

tonyd256:
This update mostly involved removing extraneous parameter labels and
lowercasing the first letter of all enums. Also, updating Curry to the
latest master commit which has the `SWIFT_VERSION=3.0` build setting.


Argo/Types/Decoded/Decoded.swift
```diff
@@ -82,8 +82,8 @@ public extension Decoded {
     - returns: A `Decoded.Failure` with a `.TypeMismatch` error constructed
                from the provided `expected` and `actual` values
   */
-  static func typeMismatch<T, U>(expected: String, actual: U) -> Decoded<T> {
-    return .Failure(.TypeMismatch(expected: expected, actual: String(actual)))
+  static func typeMismatch<T, U>(_ expected: String, actual: U) -> Decoded<T> {
```
gfontenot: I actually liked the external label here. Read much nicer IMO

ArgoTests/JSON/JSONFileReader.swift
```diff
@@ -1,13 +1,13 @@
 import Foundation
 
-func JSONFromFile(file: String) -> AnyObject? {
+func JSONFromFile(_ file: String) -> AnyObject? {
```
gfontenot: Should we rename this to `json(fromFile:)`?

ArgoTests/JSON/JSONFileReader.swift
```diff
@@ -1,13 +1,13 @@
...
   return Bundle(for: JSONFileReader.self).pathForResource(file, ofType: "json")
     .flatMap { URL(fileURLWithPath: $0) }
     .flatMap { try? Data(contentsOf: $0) }
     .flatMap(JSONObjectWithData)
 }
 
-private func JSONObjectWithData(data: Data) -> AnyObject? {
+private func JSONObjectWithData(_ data: Data) -> AnyObject? {
```
gfontenot: Ditto here, maybe `json(fromData:)` is a better name?

Argo/Types/Decoded/Decoded.swift
tonyd256: ya I agree ... I think the Swift converter in the latest Xcode switched to doing this for you so it wouldn't break changes else where ... I think this got switched because we already ran the converter with the first beta.

I'll change it back. Comment on any others you feel were better the other way.

ArgoTests/Models/URL.swift
```diff
@@ -0,0 +1,14 @@
+import Argo
+import Foundation
+
+extension URL: Decodable {
+  public typealias DecodedType = URL
```
gfontenot: Is this actually needed? AFAIK we haven't tried removing it since `static` was added as a synonym for `final class`.

ArgoTests/Tests/DecodedTests.swift
```diff
@@ -88,62 +88,62 @@ class DecodedTests: XCTestCase {
...
+    case .success: XCTAssert(result.description == successUser.description)
     default: XCTFail("Unexpected Case Occurred")
     }
   }
 }
 
 private struct Dummy: Decodable {
   static func decode(_ json: JSON) -> Decoded<Dummy> {
-    return .Failure(.Custom("My Custom Error"))
+    return .failure(.custom("My Custom Error"))
```
gfontenot: Maybe we could use the `.customError` method here?

Argo/Types/Decoded/Decoded.swift
gfontenot: The other two seemed fine with the ignored label, this is the only one that I think I preferred the other way.

Argo/Types/JSON.swift
```diff
@@ -2,12 +2,12 @@ import Foundation
 
 /// A type safe representation of JSON.
 public enum JSON {
-  case Object([Swift.String: JSON])
-  case Array([JSON])
-  case String(Swift.String)
-  case Number(NSNumber)
-  case Bool(Swift.Bool)
-  case Null
+  case object([Swift.String: JSON])
```
jakecraige: We should be able to remove the `Swift.` namespace's now. What do you think?

gfontenot: 👍

PR #394, opened 2016-07-22
Set the swift version at the project level

gfontenot:
This makes sure we're using a consistent swift version between targets
without having to modify them individually.


tonyd256:
👍

PR #395, opened 2016-07-22
[RFC] Reveal Runes as a hard dependency

gfontenot:
A while back, we used git submodules and linking trickery to hide the fact
that Argo uses Runes as a dependency for declaring monadic operator
definitions. This seemed like a good idea at the time, and has _mostly_ worked
well, but for Swift Package Manager we can't actually use the same technique
to link in Runes.swift

Additionally, using submodules in this way actually caused some subtle issues
for users:

 - If CocoaPods users wanted to specify a git branch or tag or commit or
   anything, they would also need to pass the `:submodules => true` flag in
   their podfile. This made us unlike basically any dependency.

 - Git Submodules subtly changed their behavior recently, which actually
   _completely broke_ Argo for CocoaPods users on git 2.9. It was later fixed,
   but we had to discover/debug/fix the problem ourselves, since we were
   apparently the only people that hit it. This is partially due to the fact
   that I aggressively update tools like git, but it's also a sign that we're
   a bit of an edge case, and that's not a super comfortable place to be. #
   Please enter the commit message for your changes. Lines starting

By declaring Runes as a hard, external dependency, we can get around these
issues, simplify our project setup, and move ourself to be in a better place
for supporting Swift Package Manager through the same mechanism.

The downside to this is that we've moved back to needing users to manually
import/link Runes into their files/apps when using Argo. I wish this wasn't
the case, but it still feels like the most reasonable way forward.

Opening this to get feedback from people before merging. ping @thoughtbot/ios

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,7 @@ build/
 Pods
 
 # Carthage
-Carthage/Build
+Carthage/
 
 # AppCode specific files
 .idea/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,0 @@
-[submodule "Carthage/Checkouts/Runes"]
-	path = Carthage/Checkouts/Runes
-	url = https://github.com/thoughtbot/Runes.git
-[submodule "Carthage/Checkouts/Curry"]
-	path = Carthage/Checkouts/Curry
-	url = https://github.com/thoughtbot/Curry.git
--- a/Argo.podspec
+++ b/Argo.podspec
@@ -10,8 +10,11 @@ Pod::Spec.new do |spec|
     'thoughtbot' => nil,
   }
   spec.social_media_url = 'http://twitter.com/thoughtbot'
-  spec.source = { :git => 'https://github.com/thoughtbot/Argo.git', :tag => "v#{spec.version}", :submodules => true }
-  spec.source_files = 'Argo/**/*.{h,swift}', 'Carthage/Checkouts/Runes/Source/Runes.swift'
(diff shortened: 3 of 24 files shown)


sharplet:
There _may_ be a way to remove the need to explicitly import Runes when using Argo. I first learned about this in the swiftpm codebase: https://github.com/apple/swift-package-manager/blob/2c588fd745b8d51fbc7b4d2511740768e0ff8c61/Sources/libc/libc.swift

```swift
// Sources/libc/libc.swift
#if os(Linux)
@_exported import Glibc
#else
@_exported import Darwin.C
#endif
```

This defines a module, `libc`, that purely exists to re-export either `Darwin.C` or `Glibc`, depending on your platform. Callers just need to `import libc`.

So what I'm wondering is, maybe Argo can put a `@_exported import Runes` somewhere and declare Runes as part of its public API?

gfontenot:
Oh interesting. Yeah, maybe as a part of our umbrella header? That sure looks like a private API, I wonder if that would cause problems for us?

sharplet:
It seems like a modular import in the umbrella header should work for Darwin platforms, and is a more supported configuration. However that doesn't help with swiftpm.

As far as I can tell the `libc` module is intended to be internal to swiftpm at this point in time. If we were to start using `@_exported import` as public API, then the feature was later removed, that would force a breaking change.

I personally feel though that the user experience benefits would be worth the risk of a future breaking change.

gfontenot:
I think what I'd like to do is to expose this as is, and then look into improving the end user experience with `@_exported`. I'd like to do some more research around using that API before committing to it.

PR #396, opened 2016-07-23
Support for decoding UInt and UInt64

gcox:
Tried to update the ```TestModelNumerics``` struct cleanly but ran into the *"Expression was too complex to be solved in reasonable time"* bug in Swift after I added the 3 required properties to the existing struct. That's why I added the ```TestModelUnsignedNumerics``` struct.


gcox:
I'm happy to redo this change on the Swift-2.3 branch if it should go there first

jshier:
You can get around the "too complex" errors by breaking the `curry` call into its own `let`. If that's not enough, you can also type the return method, as explained [here](https://github.com/thoughtbot/Argo/pull/356#discussion_r71237811).

gfontenot:
Neat, thanks! I think we could merge this into master and then I'll cherry-pick it into a new release for 2.3 since there's no reason _not_ to backport it.

gfontenot:
I worked around the too complex issues and merged this into master. I'm also cutting 3.1 with this change in it.

PR #399, opened 2016-08-01
Documentation for common gotchas

gfontenot:
This is a continuation of #356

### DIFF
--- a/Documentation/Compilation-Errors.md
+++ b/Documentation/Compilation-Errors.md
@@ -50,6 +50,23 @@ defining `tmp` we used the map operator `create <^>` but in the return
 statement we used the applicative operator `tmp <*>`. The map operator is
 _only_ used with the fully curried `init`.
 
+If either of these methods aren't doing the trick, or if you'd prefer to avoid
+the temporary variables you can help the compiler by adding more type
+information with `as` statements:
+
+```swift
+struct Model: Decodable {
+  let a: String
+  let b: Int
+
+  static func decode(json: JSON) -> Decoded<Model> {
+    return curry(Model.init)
+      <^> json <| "key1" as Decoded<String> // Note that these types correspond
+      <*> json <| "key2" as Decoded<Int>    // to the expected type above
+  }
+}
+```
+
 ## Incorrectly decoded fields
 
 When you have not correctly decoded all fields of your model in the `decode`
--- a/Documentation/Compilation-Errors.md
+++ b/Documentation/Compilation-Errors.md
@@ -19,11 +19,11 @@ Fortunately this is easy to solve. A first step is to store
 `curry(Model.init)` in a variable instead of using it directly:
 
 ```swift
-static func decode(j: JSON) -> Decoded<Model> {
+static func decode(json: JSON) -> Decoded<Model> {
   let create = curry(Model.init)
   return create
-    <^> j <| "key1"
-    <*> j <| "key2"
+    <^> json <| "key1"
+    <*> json <| "key2"
     ...
 ```
 
@@ -32,15 +32,15 @@ issues, the next step is to pick a midpoint in your decoding chain and set an
 intermediate variable:
 
 ```swift
-static func decode(j: JSON) -> Decoded<Model> {
+static func decode(json: JSON) -> Decoded<Model> {
   let create = curry(Model.init)
   let tmp = create
-    <^> j <| "key_a_1"
-    <*> j <| "key_a_2"
+    <^> json <| "key_a_1"
+    <*> json <| "key_a_2"
     ...
   return tmp
-    <*> j <| "key_b_1"
-    <*> j <| "key_b_2"
+    <*> json <| "key_b_1"
+    <*> json <| "key_b_2"
     ...
 ```
 
@@ -90,18 +90,18 @@ likely means that you are using `<*>` when you should be using `<^>`:
 
 ```swift
 // Correct
-static func decode(j: JSON) -> Decoded<Model> {
+static func decode(json: JSON) -> Decoded<Model> {
   return curry(Model.init)
-    <^> j <| "key1" // <-- Good
-    <*> j <| "key2"
+    <^> json <| "key1" // <-- Good
+    <*> json <| "key2"
     ...
 }
 
 // Incorrect
-static func decode(j: JSON) -> Decoded<Model> {
+static func decode(json: JSON) -> Decoded<Model> {
   return curry(Model.init)
-    <*> j <| "key1" // <-- Bad
-    <*> j <| "key2"
+    <*> json <| "key1" // <-- Bad
+    <*> json <| "key2"
     ...
 }
 ```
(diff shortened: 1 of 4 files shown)


gfontenot:
ping @thoughtbot/ios @thoughtbot/copy-editing

gfontenot:
also ping @mbrandonw in case he has any thoughts

Documentation/Compilation-Errors.md
```diff
@@ -0,0 +1,134 @@
+# Compilation errors
+
+Swift’s compile errors can be mystifying if you are not intimately familiar with
```
gabebw: "compile errors", or "compilation errors"? "Compilation errors" sounds better to me.

Documentation/Compilation-Errors.md
```diff
@@ -0,0 +1,134 @@
...
+the internals of Argo. Here we describe some of the most common errors and how
+to fix them.
+
+## Complex expressions
+
+Considering that the implementation of `decode` on a model consists of fully
```
gabebw: s/Considering that/since

Documentation/Compilation-Errors.md
```diff
@@ -0,0 +1,134 @@
...
+currying a method and composing it with many small pieces, each of which may be
+the composition of pieces, it is not surprising that Swift sometimes has
```
gabebw: s/the composition of pieces/composed of many other pieces/

Documentation/Compilation-Errors.md
```diff
@@ -0,0 +1,134 @@
...
+    <^> j <| "key_a_1"
+    <*> j <| "key_a_2"
+    ...
+  return tmp
+    <*> j <| "key_b_1"
+    <*> j <| "key_b_2"
+    ...
+```
+
+This should almost surely fix all of your complex expression issues, but if not,
```
gabebw: "should almost surely" sounds alternately very confident and very uncertain. How about: "This will almost always fix...".

Documentation/Compilation-Errors.md
```diff
@@ -0,0 +1,134 @@
...
+ * You incorrectly decoded the type of a field, e.g. you used `j <|
+   "optional_key"` when you meant `j <|? "optional_key"`.
+
+The best way forward is to re-examine all of the fields in your model and make
+sure they are accounted for in `decode` and that types match up.
+
+## Misuse of functional operators
+
+When you have misused the map `<^>` or applicative `<*>` operators you will see
+errors that mention the `Decoded` type. One such kind of error is:
```
gabebw: One such error...

Documentation/Compilation-Errors.md
```diff
@@ -0,0 +1,134 @@
...
+
+> No 'curry' candidates produce the expected contextual result type 'Decoded<_
+> -> _ -> _ -> C>'
+
+> Cannot convert value of type 'A -> B -> C' to type 'Decoded<_ -> _ -> C>'
+
+The key difference in these errors from the previous is the mention of
```
gabebw: "the previous" meaning "Incorrectly decoded fields"? I'd explicitly mention the section to look at in this document instead of saying "the previous".

Documentation/README.md
```diff
@@ -24,3 +24,8 @@ can be.
 - Interacting with the `JSON` enum // TODO
 - Writing your own custom parser // TODO
 - More complex parsers // TODO
+
+## Common Gotchas
```
gabebw: 🎉 Thank you for linking to these from the main README.

Documentation/Compilation-Errors.md
```diff
@@ -0,0 +1,134 @@
...
+former happens when you have stored `curry(Model.init)` in a variable instead of
+returning directly. See the discussion on complex expressions for more
+information on why one would do that.
+
+There are a few ways in which you may not have correctly decoded your model:
+
+ * You are missing a `j <| "key"` for a field in your model.
+ * You incorrectly decoded the type of a field, e.g. you used `j <|
+   "optional_key"` when you meant `j <|? "optional_key"`.
+
```
jakecraige: I think there's another case here where you're decoding a type that doesn't implement `Decodable`? I'm not sure if it's this exact error or not, but I know that's something I've ran across where the compiler error wasn't clear so it could be nice to mention it.

jakecraige:
Wow this is awesome. This would have been immensely helpful when I was getting started with Argo. Great work!

Documentation/Compilation-Errors.md
gfontenot: 👍

Documentation/Compilation-Errors.md
gfontenot: 👍

Documentation/Compilation-Errors.md
gfontenot: 👍

Documentation/Compilation-Errors.md
gfontenot: 👍

Documentation/README.md
gfontenot: This is actually the documentation README. Main -> Documentation -> Individual

Documentation/Compilation-Errors.md
gfontenot: I think it'd probably be useful to merge this in now without trying to track all of these down, and then add more in later as we encounter them.

Documentation/Compilation-Errors.md
gfontenot: 👍

Documentation/Compilation-Errors.md
gfontenot: 👍

Documentation/Compilation-Errors.md
```diff
@@ -0,0 +1,138 @@
...
+    <*> j <| "key_b_2"
+    ...
+```
+
+This almost always fix all of your complex expression issues, but if not, be
+prepared to try one more intermediate variable. Take careful note that in
+defining `tmp` we used the map operator `create <^>` but in the return
+statement we used the applicative operator `tmp <*>`. The map operator is
+_only_ used with the fully curried `init`.
+
```
tonyd256: If either of these methods isn't doing the trick or if the user simply prefers the look of assigning types to all those lines, One can achieve the limit of curry without having to separate the `curry(Model.init)` from the `return`.

```swift
static func decode(j: JSON) -> Decoded<Model> {
   return curry(Model.init)
     <^> j <| "key1" as Decoded<String>
     <*> j <| "key2" as Decoded <Int>
     ...
 ```

File: Documentation/Compilation-Errors.md [hunk 10]
```diff
@@ -0,0 +1,138 @@
...
+    ...
+```
+
+Many times this can fix the problem. If you still have complex expression
+issues, the next step is to pick a midpoint in your decoding chain and set an
+intermediate variable:
+
+```swift
+static func decode(j: JSON) -> Decoded<Model> {
+  let create = Curry(Model.init)
```
tonyd256: lower case `curry`?

tonyd256: actually everywhere it's capital so find/replace

Documentation/README.md
gabebw: Well, still.

tonyd256:
👍  after Curry -> curry

gabebw:
👍  from me.

Documentation/Compilation-Errors.md
jakecraige: I agree that we don't need to track all of them down right now, but this is one I have personally encountered many times over, so I think it's worth mentioning. 

I don't think it's blocking this PR to not have it, but I think it's something we definitely should include. If not in this PR, in one after.

gfontenot: I'm just not sure what that error looks like, and don't really have time to figure it out. If you know the error I'll happily add it.

jakecraige: Sure thing, here's an example and error:

screen shot 2016-08-02 at 15 02 00

gfontenot: Added

PR #411, opened 2016-09-09
Add Carthage directories to framework search paths

gfontenot:
When building with Carthage locally, Argo uses the local checkouts of the
dependencies to build and link against. However, when using Argo as a
dependency, those checkouts don't exist. This means that when Argo tries to
build, it can't find Runes and so the build fails.

Unfortunately, if we move to a more carthage centric workflow where we use the
built binaries to link against, that ends up breaking the workflow for users
that want to use Git Submodules for their dependency management (because they
won't have those frameworks in those locations).

We can work around this while still allowing users to use submodules by adding
the build directories to the framework search paths _without_ explicitly using
those built frameworks in our project. This means that Carthage users will be
able to see the Runes dependency, and submodules users will be able to use the
project and implicit dependencies.

The only downside here is that this might result in warnings for submodules
users. I think that's a reasonable trade-off.

### DIFF
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -1179,6 +1179,10 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(inherited)",
+					"$(PROJECT_DIR)/Carthage/Build/tvOS",
+				);
 				GCC_NO_COMMON_BLOCKS = YES;
 				GCC_TREAT_WARNINGS_AS_ERRORS = YES;
 				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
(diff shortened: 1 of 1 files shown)


mdiep:
You might be helped by [this PR](https://github.com/Carthage/Carthage/pull/1421), which will be included in the next Carthage release sometime this weekend. Otherwise you could use `--use-submodules` to make your dependencies available.

gfontenot:
Oh, killer. Yeah I think we could just rely on this instead of adding this search path. Although I guess that could be subverted if we shipped pre-built binaries for Runes?

mdiep:
Pre-built binaries for Runes shouldn't affect how Argo itself is built if you're using a workspace to build Argo that has Runes as a dependency. Consumers of Argo could still benefit from pre-built binaries of both Argo and Runes, but Argo won't try to use the copy from `Carthage/Build/` unless that's the only setup that you support.

gfontenot:
Maybe I'm misunderstanding? If Argo relies on the checkout of Runes to build (like it does currently), and we supply a pre-built binary of Runes (which we currently don't), wouldn't users end up pulling down a binary of Runes, skipping the checkout, and therefore breaking the build? (If we don't modify the project or merge this PR, I mean).

mdiep:
Oh, you're right. You'd have to add Runes as a submodule. 😕 

Carthage should always checkout dependencies—even if a pre-built binary is available. I opened a Carthage issue for this: https://github.com/Carthage/Carthage/issues/1466

gfontenot:
👍

For now though, I think since we're in control of both of these dependencies we can safely rely on the (soon-to-be-new) default behavior of Carthage instead of doing anything in this PR. If we ship pre-built binaries in the future we'll need to revisit.

gfontenot:
Oh, Carthage 0.18 is actually already released! Neato.

PR #413, opened 2016-09-16
A missing key in an embedded type should be a failure

tonyd256:
This addresses the issue that an embedded object that is optionally decoded, as
the `Location` object is in the `LocationPost` object in these changes, should
fail to decode if the key exists but the embedded json fails to parse for any
reason. Right now there is a bug that a missing key failure in the embedded
object will cause the parent object to have a `.success(.none)` for its child
instead of the correct `.failure`.

### DIFF
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -107,6 +107,15 @@
 		EA47BB591AFC5E65002D2CCD /* user_without_key.json in Resources */ = {isa = PBXBuildFile; fileRef = EA47BB581AFC5E65002D2CCD /* user_without_key.json */; };
 		EA47BB5A1AFC5E65002D2CCD /* user_without_key.json in Resources */ = {isa = PBXBuildFile; fileRef = EA47BB581AFC5E65002D2CCD /* user_without_key.json */; };
 		EA4EAF7319DD96330036AE0D /* types_fail_embedded.json in Resources */ = {isa = PBXBuildFile; fileRef = EA4EAF7219DD96330036AE0D /* types_fail_embedded.json */; };
+		EA6600ED1D8C648F0032103D /* EmbeddedDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA6600EC1D8C648F0032103D /* EmbeddedDecodingTests.swift */; };
+		EA6600EE1D8C648F0032103D /* EmbeddedDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA6600EC1D8C648F0032103D /* EmbeddedDecodingTests.swift */; };
+		EA6600EF1D8C648F0032103D /* EmbeddedDecodingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA6600EC1D8C648F0032103D /* EmbeddedDecodingTests.swift */; };
+		EA6600F11D8C651E0032103D /* location_post.json in Resources */ = {isa = PBXBuildFile; fileRef = EA6600F01D8C651E0032103D /* location_post.json */; };
+		EA6600F21D8C651E0032103D /* location_post.json in Resources */ = {isa = PBXBuildFile; fileRef = EA6600F01D8C651E0032103D /* location_post.json */; };
+		EA6600F31D8C651E0032103D /* location_post.json in Resources */ = {isa = PBXBuildFile; fileRef = EA6600F01D8C651E0032103D /* location_post.json */; };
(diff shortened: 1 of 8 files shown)


pteasima:
@tonyd256 good stuff. I agree that from the standpoint of the whole decode operation, this is a `.typeMismatch` error.

Im a little dissapointed that we are losing the original error information. If the mapping failed at the top level (the one that gets reported to client code), we would know the true cause (e.g. `.missingKey("foo")`). But here, with nested mappings, we are losing that information. Its logically ok to lose the information, because we are moving to a higher level of abstraction. But from a usefulness standpoint, I would be in favor of something like: `case typeMismatch(expected: String, actual: String, underlyingError: DecodeError?)`. (I think the enum would have to be made indirect.) Of course we could end up with a long chain of such nested errors, but I think the models we parse never have too many layers and it would actually be nice to see a "path" to the source of the problem.

On the other hand, this is already an edge case and maybe we dont wanna polute the `.typeMismatch` error case with this third parameter.

tonyd256:
@pteasima I agree that it isn't great to lose that info. It actually might be better to use `.multiple` from #388 ... We could hold onto the `missingKey` error but also add the `typemismatch` error we're generating here. That would allow for keeping the lower level error as well as the higher abstraction error. Win Win?

pteasima:
Using `.multiple` doesnt feel right - these arent 2 distinct errors, just two interpretations of the same error (Child object interprets it as `.missingKey("foo")` and Parent object interprets it as `.typeMismatch(expected: "Child",...`. Hope this makes sense, I must admit I havent read #388 thoroughly.

However, since I dont have a better proposal, Im am absolutely OK with using `.multiple`. In real apps, there is usually no error handling logic for mapping errors, just printing for debugging purposes. So I feel like we dont have to do this "correctly", we just have to do it in a way that is useful. And using `.multiple` definitely does that. 👍  Let's not do anything too complicated.

tonyd256:
@pteasima Ya I hear that. I think there is a "quick" solution that makes at least the docs in line with how it should work. But maybe there is a "correct" way which might look something like you've posted above with `underlyingError`. If this was a critical bug fix I would say let's do "quick" for now and worry about correctness later. But I think this will come out with the next major version of Argo so maybe we just do the "correct" way now.

@gfontenot any thoughts here?

gfontenot:
I'd vote for the "correct" way, but we also need to have some urgency around getting a Swift 3.0 version of Argo shipped.

tonyd256:
@gfontenot do you think the `underlyingError` method above is the right way? It probably wouldn't be too hard to do that but would be a bit of a special case and since we can't have defaults on enums it would require us to have that parameter in all the places (except where we use the helper function of course).

gfontenot:
I haven't thought about it enough to really say for sure.

gfontenot:
I guess I'll default to my normal question: What does Aeson do in this case?

tonyd256:
For reference, Aeson only returns the embedded error.

```haskell
import Data.Aeson
import Data.Aeson.Types
import Data.ByteString.Char8 (pack)

data Post = Post
  { author :: Maybe User
  , text :: String
  } deriving Show
  
data User = User
  { name :: String
  , id :: Int
  } deriving Show
  
instance FromJSON User where
  parseJSON (Object o) = User
    <$> o .: "name"
    <*> o .: "id"
  
  parseJSON x = typeMismatch "User" x
  

instance FromJSON Post where
  parseJSON (Object o) = Post
    <$> o .:? "user"
    <*> o .: "text"
  
  parseJSON x = typeMismatch "Post" x
```
then
```haskell
eitherDecodeStrict (pack "{\"name\":\"Joe\"}") :: Either String User
-- Left "key \"id\" not present"

eitherDecodeStrict (pack "{\"text\":\"hello\",\"user\":{\"name\":\"Joe\"}}") :: Either String Post
-- Left "key \"id\" not present"

eitherDecodeStrict (pack "{\"text\":\"hello\"}") :: Either String Post
-- Right (Post {author = Nothing, text = "hello"})
```

I dug in a bit and it seems that having `.optional` wrap the entire result is the wrong way. We should succeed with a `.none` only when that key is missing and not when the embedded type fails to decode. So I reworked Argo to align more with Aeson in this regard. You can check out how Aeson implements `.:?` here: https://github.com/bos/aeson/blob/master/Data/Aeson/Types/FromJSON.hs#L670

I pushed new code and I think this code is actually the most "correct" especially if we consider Aeson as our inspiration.

jshier:
Just got bit by this today. Any update on landing this or the other PRs for an official Swift 3 release?

Issue #412, opened 2016-09-14
no such module 'Runes'


edwardaux:
I have Argo embedded within my existing [project](https://github.com/edwardaux/Ogra) (and have had for ages). I'm finally getting around to updating to Swift 3.

I updated my project's `Cartfile`to point to Argo _master_ (instead of _3.1.0_):

```
github "thoughtbot/Argo" "master"
```

and run `carthage update --use-submodules`. This successfully fetches Argo and Runes. As you can see from the log below, Runes compiles OK and I can see the artifacts in the `Carthage/Build/XXX` folders.  However, when it tried to build Argo, I get an error saying that it can't find the Runes module.

There is a mention of a similar problem in #407 (which I don't think is related to the PR... just a similar problem). 

I've tried a bunch of carthage combos (bootstrap, build, update, et al), but just can't get a working build.  

I'm still digging through the logs, but any thoughts would be great.
### Error Messages

```
*** xcodebuild output can be found in /var/folders/p6/xcv59nqd4wv0y6hwpm3q16m40000gn/T/carthage-xcodebuild.7gtojh.log
*** Building scheme "Runes-tvOS" in Runes.xcodeproj
*** Building scheme "Runes-Mac" in Runes.xcodeproj
*** Building scheme "Runes-iOS" in Runes.xcodeproj
*** Building scheme "Runes-watchOS" in Runes.xcodeproj
*** Building scheme "Argo-watchOS" in Argo.xcworkspace
** BUILD FAILED **


The following build commands failed:
    CompileSwift normal armv7k
    CompileSwiftSources normal armv7k com.apple.xcode.tools.swift.compiler
    Ld /Users/edwardsc/Library/Developer/Xcode/DerivedData/Argo-asybjaazisryjmeuwabufbberqcz/Build/Products/Release-watchos/Argo.framework/Argo normal armv7k
    GenerateDSYMFile /Users/edwardsc/Library/Developer/Xcode/DerivedData/Argo-asybjaazisryjmeuwabufbberqcz/Build/Products/Release-watchos/Argo.framework.dSYM /Users/edwardsc/Library/Developer/Xcode/DerivedData/Argo-asybjaazisryjmeuwabufbberqcz/Build/Products/Release-watchos/Argo.framework/Argo
(4 failures)
/Users/edwardsc/dev/Ogra/Carthage/Checkouts/Argo/Argo/Types/Decoded/Applicative.swift:1:8: error: no such module 'Runes'
[...]

PR #414, opened 2016-09-24
Try to fix how Runes and Curry are included so things build.

sgoodwin:
I think this should fix the issue with module not found errors for apps using Argo.


gfontenot:
Hard to tell from the diff: Is this just adding the Carthage build folder to the framework search paths?

My concern in "fixing" this like this is that this is _only_ a problem when using Argo with Carthage and submodules. No other configuration has any issues. Once the checkout dir is symlinked for submodules the way it is for normal checkouts this issue will dissapear.

sgoodwin:
Oh, so you still want to do Runes and Curry as submodules?

I went through and pointed all those framework references in the workspace file to their places in ```Carthage/Build``` so they'd build fine. Even Argo by itself wouldn't build for all targets before that. If this doesn't help, then no worries.

gfontenot:
No, sorry. Argo itself should be building fine with Carthage assuming you're using 0.18+. The issue that we've been talking about in #412 is limited to users using `carthage --use-submodules`. That's the only time (AFAIK) that building is broken. So since that's a bug in Carthage, I was wondering if it was something we should be working around on our end.

sgoodwin:
For our project, we have to add '''--platform iOS''' to keep it from trying to build for WatchKit and such and fail.

mdiep:
> So since that's a bug in Carthage

There's a PR open for that, btw. (https://github.com/Carthage/Carthage/pull/1493)

Hopefully we can get it reviewed, merged, and released in the next week.

gfontenot:
@sgoodwin that sounds like a new issue that I haven't seen reported yet. Mind opening an issue so we can discuss?

@mdiep 🎉

sgoodwin:
Will do later today.

PR #424, opened 2016-11-02
Add Swift Package Manager support

gfontenot:
This adds support for building with Swift Package Manager. It's largely
just project reorganization and the addition of a `Package.swift` file.

Note that since Swift Package Manager still doesn't support test-only
dependencies, we're not able to get `swift test` to work, which is a
real bummer. However, I've gone ahead and moved the test code into a
folder structure that should work for `swift test` once we can use it.


gfontenot:
ping @thoughtbot/ios 

I'd like to get this in alongside the fix from #425 so that we can use Argo with SPM.

PR #425, opened 2016-11-02
Updates for Xcode 8.1, Swift 3.0.1

gfontenot:
See individual commits for more info.


sharplet:
So I thought this crash seemed a bit weird, and I had a play around. I found that the two literals in the tests weren't actually resolving to numbers of the expected type: https://github.com/thoughtbot/Argo/commit/277d4918128e19d6a882dacb26ecabb730107798. Double literals will be doubles without any other type annotations, and likewise integer literals prefer Int over UInt.

This snippet reproduces the crash:

```swift
import Foundation
let a: Any = 1.1
let b = a as! NSNumber
let c = b as Float // crash
```

But then, this _also_ crashes:

```swift
import Foundation
let a: Any = 1000 as UInt
let b = a as! NSNumber
let c = b as Int // crash
```

This seems like a bug to me, and it's definitely a regression. While using `NSNumber`'s `*Value` properties cast the value safely, I don't think we should _have_ to do it that way. Having said that, I think it makes sense to fix this crash ASAP.

Before this is merged we should change all the other places where we're using `as` to cast an `NSNumber`, because they're all potentially broken.

gfontenot:
I asked Joe Groff about this:

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr">It seems like this: <a href="https://t.co/j20ee1uN9W">https://t.co/j20ee1uN9W</a><br><br>might have been an accidental side effect of this? <a href="https://t.co/7lZr4c9NZO">https://t.co/7lZr4c9NZO</a><br><br>/cc <a href="https://twitter.com/jckarter">@jckarter</a></p>&mdash; Gordon Fontenot (@GFontenot) <a href="https://twitter.com/GFontenot/status/793611396487405568">November 2, 2016</a></blockquote> <script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr"><a href="https://twitter.com/GFontenot">@GFontenot</a> The intended behavior is that a bridged NSNumber only bridges back to its original Swift type. We probably shouldn&#39;t allow `as`</p>&mdash; Joe Groff (@jckarter) <a href="https://twitter.com/jckarter/status/793614752068931584">November 2, 2016</a></blockquote> <script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr"><a href="https://twitter.com/jckarter">@jckarter</a> would you recommend that we remove all of those casts and use the methods instead? Seems like that might be safer long term?</p>&mdash; Gordon Fontenot (@GFontenot) <a href="https://twitter.com/GFontenot/status/793627901807570944">November 2, 2016</a></blockquote> <script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr"><a href="https://twitter.com/GFontenot">@GFontenot</a> Yeah, the properties would be safer, and probably more efficient.</p>&mdash; Joe Groff (@jckarter) <a href="https://twitter.com/jckarter/status/793628729742209024">November 2, 2016</a></blockquote> <script async src="//platform.twitter.com/widgets.js" charset="utf-8"></script>

sharplet:
Seems legit 👍 What do you think about including https://github.com/thoughtbot/Argo/commit/277d4918128e19d6a882dacb26ecabb730107798?

gfontenot:
I don't think we actually need that (these tests are now green), and I don't think we want to promote that as an acceptable solution for end users. I'm actually happy that the tests caught this specific crash here, and would like to have similar crashes caught in the future.

sharplet:
Sure, catching crashes is good. I was mostly thinking it would make sense to have the `"float"` key in the tests actually be bridging a `Float`, rather than a `Double` (for example). Testing that a JSON number can be safely decoded as any Swift numeric type feels like a slightly different test?

gfontenot:
I don't think so? IMO, this test is testing exactly the behavior we want, which is that a native swift dictionary can be parsed via Argo alongside JSON and plists. I wouldn't want to need to add that type info to my swift dictionary in order to get it to work with Argo, so I don't think we should be adding the type info in our tests.

sharplet:
> I wouldn't want to need to add that type info to my swift dictionary in order to get it to work with Argo

I totally agree with you on this when it comes to literals. :+1: I suppose the behaviour I'm describing is more likely if a `Float`, `CGFloat`, `UInt`, etc. is passed in as a parameter or some other place where the type annotation is explicit.

tonyd256:
👍  we should push this asap with a version bump

gfontenot:
There's now [a bug tracking this on Jira](https://bugs.swift.org/browse/SR-3117)

PR #428, opened 2016-11-06
Improve the way we detect Boolean values

gfontenot:
Previously, we were relying on functionality found in CoreFoundation in
order to determine if an NSNumber instance was representing an
underlying Boolean value. This worked well, but unfortunately
CoreFoundation isn't available on Linux, which means that Argo would
never be able to compile. We'd like Argo to be as widely availably as
possible, so we need to find another solution.

Luckily, it looks like we can use `type(of:)` to determine this. That
function ships with Swift itself and so _should_ mean that we're now
able to compile on Linux without any behavioral change.

I added another test to ensure that our decoding is operating as we'd
expect. The previous test seemed like it was vulnerable to false
positives.

### DIFF
--- a/Sources/Argo/Extensions/NSNumber.swift
+++ b/Sources/Argo/Extensions/NSNumber.swift
@@ -2,6 +2,6 @@ import Foundation
 
 extension NSNumber {
   var isBool: Bool {
-    return CFBooleanGetTypeID() == CFGetTypeID(self)
+    return type(of: self) == type(of: NSNumber(value: true))
   }
 }
--- a/Tests/ArgoTests/Tests/TypeTests.swift
+++ b/Tests/ArgoTests/Tests/TypeTests.swift
@@ -41,4 +41,10 @@ class TypeTests: XCTestCase {
     XCTAssert(bools?.bool == true)
     XCTAssert(bools?.number == true)
   }
+
+  func testBooleanIdentification() {
+    let j = json(fromFile: "booleans").map(JSON.init)!
+    let boolean: JSON? = (j <| "realBool").value
+    XCTAssert(boolean == .bool(true))
+  }
 }


sharplet:
Do we have a way to verify this behaves as expected on Linux?

Sources/Argo/Extensions/NSNumber.swift
```diff
@@ -2,6 +2,6 @@ import Foundation
 
 extension NSNumber {
   var isBool: Bool {
-    return CFBooleanGetTypeID() == CFGetTypeID(self)
+    return type(of: self) == type(of: NSNumber(value: true))
```
tonyd256: do we need to construct `NSNumber` to check if the type is `Bool`? could we simply compare it to `Bool` or `type(of: true)`?

gfontenot: No, it needs to check to see if it's an NSNumber:

```
  2> type(of: true)
$R0: Bool.Type = Bool
  3> type(of: NSNumber(value: true))
$R1: NSNumber.Type = 0x00000001005f5040
  4> type(of: NSNumber(value: 5))
$R2: NSNumber.Type = 0x00000001005f5068
```

gfontenot:
> Do we have a way to verify this behaves as expected on Linux?

You're right, we should confirm this before merging.

ikesyo:
> Previously, we were relying on functionality found in CoreFoundation in
order to determine if an NSNumber instance was representing an
underlying Boolean value. This worked well, but unfortunately
CoreFoundation isn't available on Linux, which means that Argo would
never be able to compile.

I think adding `import CoreFoundation` should fix the issue.

ikesyo:
Submitted #430.

gfontenot:
Ha, sure I guess that makes sense.

I _am_ tempted to use this solution though, as it means we no longer depend on CoreFoundation at all, which always made me nervous anyway.

nickcharlton:
I've just tested this on Ubuntu 16.04 (Xenial) with Swift 3.0.1-RELEASE and it compiles fine.

PR #429, opened 2016-11-06
Improve CircleCI setup

gfontenot:
This is still going to fail because (as per usual) Circle is broken
(they have duplicate simulators installed, which makes it impossible to
refer to them simply by name), but once they fix their stuff this should
be much nicer.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,19 +1,53 @@
-#!/usr/bin/env sh
+#!/bin/bash
+#
+# Runs tests via xcodebuild, formatting results with xcpretty.
+#
+# Usage: bin/test [options]
+#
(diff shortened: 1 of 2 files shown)


sharplet:
FWIW, I believe Circle will now autogenerate a pretty sensible xcodebuild + xcpretty command, including the JUnit reporting stuff, using their inference features.

Not sure if we want to rely on that, but it would simplify `circle.yml` further to not override the test action.

gfontenot:
Yeah, I'd still rather not rely on Circle for this. I use the `bin/test` script myself to verify that things are OK, so I'm more interested in improving that and using it consistently.

PR #465, opened 2017-06-04
Re-export Runes and redeclare its operators

sharplet:
(For some context, see https://github.com/ReactiveCocoa/ReactiveCocoa/issues/3450.)

Unfortunately, both when using `@_exported` or importing via the umbrella header, operator declarations still don't show up in the consuming module. This is particularly frustrating for Runes, given that its whole job is exporting operators.

The workaround is to redeclare all the same operators, reusing the same precedence groups. This is kind of annoying, and could result in some minor maintenance issues down the line as new operators are introduced, but ultimately _it works_. Which is cool.

Given that `@_exported` is an undocumented feature, I've done my best not to rely on it when we don't have to. That is, when building as a framework, we instead use modular imports in the bridging header to export dependencies. So it's only when building via swiftpm that `@_exported` is used.

### DIFF
--- a/Argo.podspec
+++ b/Argo.podspec
@@ -11,7 +11,9 @@ Pod::Spec.new do |spec|
   }
   spec.social_media_url = 'http://twitter.com/thoughtbot'
   spec.source = { :git => 'https://github.com/thoughtbot/Argo.git', :tag => "v#{spec.version}" }
-  spec.source_files = 'Sources/**/*.{h,swift}'
+  spec.source_files = 'Sources/**/*.swift', 'Resources/Argo.h'
+  spec.module_map = 'Sources/Argo/include/module.modulemap'
+  spec.public_header_files = 'Resources/Argo.h'
 
   spec.dependency 'Runes', '>= 4.0.0'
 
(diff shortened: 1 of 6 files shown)


sharplet:
I just pushed a fix to the podspec: frustratingly, CocoaPods ignores our umbrella header by default, generating its own for us. It generates a header with the suffix `-umbrella`, which is non-standard, and I think only works because they also generate a custom module map.

Leaving aside whether this change should be made to Runes itself, in CocoaPods environments Argo's umbrella is now effectively renamed: from `Argo-umbrella.h` to `Argo.h`. I suppose this is technically a breaking change? This whole thing is way more complicated than I'd hoped.

sharplet:
Having said that, we don't actually support Objective-C. So I think this is fine.

gfontenot:
I'd love this if it was more supported, but given the private nature of `@_exported` and needing to re-export operator definitions, I'm not sure I'd want to merge this. Is there a compelling reason (besides convenience) for moving in this direction?

sharplet:
> Is there a compelling reason (besides convenience) for moving in this direction?

Nope, I think it's just one of those things that would be kinda nice. One less "what's the deal with this confusing compiler err... oh yeah I forgot to import something".

gfontenot:
I think that I'd be OK with this if we didn't have to re-declare the operators from Runes. Since we do, I think I'm 👎 on this.

sharplet:
Hopefully we get proper support for this at some point.

PR #467, opened 2017-06-12
Fix tests for Swift 4

gfontenot:
This updates the test suite for Swift 4 (basically just disambiguating `Argo.Decodable` from `Swift.Decodable`) and updates the test suite/package definition so that we can run the tests using `swift test`.

Note that this also _removes_ support for running the tests from Xcode. See 422be95 for the reasoning behind this change.

### DIFF
--- a/Tests/ArgoTests/Models/Booleans.swift
+++ b/Tests/ArgoTests/Models/Booleans.swift
@@ -2,7 +2,7 @@ import Argo
 import Curry
 import Runes
 
-struct Booleans: Decodable {
+struct Booleans: Argo.Decodable {
   let bool: Bool
   let number: Bool
 
--- a/Tests/ArgoTests/Models/Comment.swift
+++ b/Tests/ArgoTests/Models/Comment.swift
@@ -8,7 +8,7 @@ struct Comment {
   let authorName: String
 }
 
-extension Comment: Decodable {
+extension Comment: Argo.Decodable {
   static func decode(_ json: JSON) -> Decoded<Comment> {
     return curry(self.init)
       <^> json <| "id"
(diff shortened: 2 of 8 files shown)


gfontenot:
CI will fail because we're explicitly telling it to fail.

jshier:
IMO it's premature to remove the ability to run tests through Xcode. Until SPM is actually integrated into Xcode, contributors will have to make their edits in Xcode and then run the command line tool to run their tests, losing all of the IDE integrations provided for testing. Your theory about eventually dropping support for CocoaPods/Carthage may turn out to be true but is likely at least a year away, especially given the rate at which SPM is working on their Xcode integration. I'd suggest keeping the Xcode tests and updating the CI integration to do the multiple testing channels so any friction is there and not on the developer.

Package.swift
```diff
@@ -1,8 +1,34 @@
+// swift-tools-version:4.0
```
mdiep: There are 3 paths for building with SwiftPM:

1. Swift 3 compiler, Swift 3 language
2. Swift 4 compiler, Swift 3 language
3. Swift 4 compiler, Swift 4 language

My understand is adding this line removes support for (1). You may want to remove this line, which I believe would allow you to build with all 3 options.

gfontenot: Removing this line actually causes `swift build` to fail because it's trying to default to the 3.2 tooling (which doesn't understand this package format). I'm still not sure what we _want_ here long term, but I'm thinking we're probably going to end up dropping 3.1 support and pushing forward with the 4.0 style in the future.

gfontenot:
Been thinking about this PR a lot, and haven't been sure what to do with it. I'm moving the SPM stuff into a new branch (gf-spm-all-in) because like @jshier said, it's probably a bit premature to push everything in that direction. For now, I'm going to refocus this PR on fixing the tests for Swift 4, which will let us move forward with Swift 4 related renaming and breaking changes.

gfontenot:
This was merged into master, not just closed. I forgot that I had a separate Swift 4 branch going.

Issue #418, opened 2016-10-12
Long compile times


mohandev:
I'm using Argo in production app, and I can see that the Swift compile time takes a lot of time in my project. Wondering if this is because of Argo & Argo based models which use a lot of generic code ? Please let me know if you have done any bench marking around this.

### Argo Version

Argo 3.1

### Dependency Manager

Carthage

tonyd256:
Most likely this is due to a large model that has some complex or large quantity of Argo decoding going on. Argo usually speeds up with smaller models or simpler decoding setups. If you paste some of your bigger models here we could help advise where a slowdown maybe occurring.

mohandev:
@tonyd256 Thanks for replying. I don't have single large model in the app, however I've around 20 smaller models with max 15 properties. So I hope this won't hurt the compile time ?

I tried to measure the compile time with "-Xfrontend -debug-time-function-bodies" and I don't see much compilation times reported Argo based models. I hope the above flag would have exposed if argo based models are at fault ?

tonyd256:
15 properties is a large enough model to slow down compile time for sure. I'd suggest trying to break up anything around 10-12 properties into smaller models. You could also try making it easier for the compiler by adding an explicit type to each line like so: `<*> j <| "name" as Decoded<String>` (might need parens)

mohandev:
@tonyd256, thanks for the suggestions. Will do that for the larger models!

dalu93:
FYI, @tonyd256 solution works quite fine as temporary fix. One of our project's model was compiling in **11568.x ms** and now, with the explicit type, it takes only **26.x ms**

tonyd256:
@dalu93 Great to hear! Unfortunately, I think this is a limitation of the Swift compiler and there isn't much Argo can do to lighten the load without sacrificing what it is. Closing this for now, but feel free to continue commenting or create a new issue for anything new!

dalu93:
Hi @tonyd256, the workaround is not working. The error I get is:

    No '<|' candidates produce the expected contextual result type 'Decoded<_>'

The configuration I have is:

    Argo 4.1.0
    Curry 3.0.0
    Runes 4.0.1

Is there a way to apply the same behavior? I'm getting again long compile times unfortunately

gfontenot:
@dalu93 you might need to add some parens around the `j <| "foo"` part of the expression. This is due to precedence changes we needed to make alongside Swift 3

zhijunsheng:
[...]

PR #470, opened 2017-07-11
Improve documentation for long compilation time

wongzigii:
Related to #418

### DIFF
--- a/Documentation/Compilation-Errors.md
+++ b/Documentation/Compilation-Errors.md
@@ -6,6 +6,8 @@ errors and how to fix them.
 
 ## Complex expressions
 
+#### Note: Complex expressions will somehow increase the build/compilation time.
+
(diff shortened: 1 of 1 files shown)


gfontenot:
Seems like we might be better served updating https://github.com/thoughtbot/Argo/blob/master/Documentation/Compilation-Errors.md#complex-expressions to mention long compile times in addition to "Expression too complex" errors.

wongzigii:
I added a note tip to the section header and deleted the outdated `as` usage. (no longer work with Swift 3.0 or later)

gfontenot:
Hey @wongzigii, thanks so much for this and I sincerely apologize for disappearing on this. I've opened #472, which is basically an alternative version of this PR. Does that one work for you? If so, I'll merge that one and close this one.

PR #486, opened 2018-04-25
Remove <|| family of operators

gfontenot:
We can now use conditional conformance to remove the need for the `<||` family
of operators, and instead make `[T]` conform directly to `Decodable` if `T:
Decodable`.

Note that this means we can _also_ do this for `[String: T: Decodable]`, which
means you can now use `<|` to decode directly to `[String: T]`.

This is the first pass of simplifying the API for Argo 5.0. I think we're
going to go _much_ further with these changes, because of how much simpler we
can make the API with the new abilities of Swift 4.1. Unfortunately, we're not
going to be able to avoid breaking changes here, but I hope that the fixes are
mostly focusing the API into a narrower set of operators.


gfontenot:
@tonyd256 and I have been talking about this day for nearly 3 years now. It's finally happening!

tonyd256:
This PR makes me so happy.

Sources/Argo/Types/StandardTypes.swift
```diff
@@ -194,40 +194,15 @@ public extension Collection where Iterator.Element: Decodable, Iterator.Element
 
     - returns: A decoded array of values
   */
-  static func decode(_ json: JSON) -> Decoded<[Iterator.Element]> {
+  public static func decode(_ json: JSON) -> Decoded<[Iterator.Element]> {
```
sharplet: `Decoded<[Element]>`? Could also be `Decoded<Array>`, which would implicitly have the same `Element` type as the extension.

gfontenot: Oh right. This is an artifact from when this was an extension on `Collection`. I'll go with `Decoded<Array>`, I dig that.

gfontenot: Made the same change for `Dictionary`.

PR #487, opened 2018-04-26
Misc test suite improvements

gfontenot:
- Prefer `XCTAssertEqual` over `XCTAssert` to improve failure messages.
 - Remove redundant `nil` testing
 - Remove force-unwrapping to avoid crashing the test suite


gfontenot:
FUN FACT: I'm fairly sure that the use of `XCTAssert` is an artifact dating back to Swift 1.x where optionals were constantly in flux and the `XCTAssertEqual` functions couldn't handle them.

Tests/ArgoTests/Tests/EquatableTests.swift
```diff
@@ -6,13 +6,13 @@ class EquatableTests: XCTestCase {
     let j = json(fromFile: "types").map(JSON.init)
     let anotherParsed = json(fromFile: "types").map(JSON.init)
 
-    XCTAssertEqual(j!, anotherParsed!)
+    XCTAssertEqual(j, anotherParsed)
```
sharplet: nice

PR #488, opened 2018-04-27
Replace <| operator with generic subscript

gfontenot:
The `<|` operator was necessary when we originally implemented Argo, but Swift
has new capabilities now that reduce the need to use a custom operator.
Additionally, the operator spelling itself actually conflicts with prior-art
that we weren't aware of when we added it. Specifically, `<|` is used in other
languages (and sometimes in Swift) as a "pipe backwards" operator, with the
type `(T -> U) -> T -> U`.

Given these concerns, as well as the additional complexity around fine-tuning
the precedence so that it works in an obvious way, I think it makes sense to
remove the operator in favor of a generic subscript (which only just became
available in Swift 4).


Sources/Argo/Functions/decode.swift
```diff
@@ -79,7 +79,7 @@ public func decode<T: Decodable>(_ object: Any) -> T? where T == T.DecodedType {
   - returns: A `Decoded<T>` value where `T` is `Decodable`
 */
 public func decode<T: Decodable>(_ dict: [String: Any], rootKey: String) -> Decoded<T> where T == T.DecodedType {
-  return JSON(dict as Any) <| rootKey
+  return JSON(dict as Any)[rootKey]
```
gfontenot: dis ugly

jshier:
This is one short step away from [dynamic member lookup](https://github.com/apple/swift-evolution/blob/master/proposals/0195-dynamic-member-lookup.md) which will be coming in Swift 4.2. Instead of `json["key"]`, it will allow `json.key`. It will be interesting to see if Argo can maintain it's error fidelity in that usage.

gfontenot:
Oh man, I didn't even think about that. Great point! I wonder if we'd be able to support chaining for nested objects in that way as well? I'll be super interested in playing with this at some point.

PR #491, opened 2018-05-27
Rename RawRepresentable to RawRepresentableTests to avoid name conflicts

RLovelett:
When I'm importing the test suite on Linux in a `LinuxMain.swift` it is complaining about a name conflict.

Presumably this was just a typo to begin with so I'm proposing this fix. Otherwise, I can just do `Argo.RawRepresentable`. Though that just seem suboptimal.


gfontenot:
🤦‍♂️ yep, this should definitely be what this is named.

Thanks!

PR #494, opened 2018-09-19
README: replace the `<|` operators with `[]` in the example

eunikolsky:
This is to follow-up the [PR][] that removed `<|` operators from the
code.

[PR]: https://github.com/thoughtbot/Argo/pull/488

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,12 +110,12 @@ struct User {
 extension User: Decodable {
   static func decode(_ json: JSON) -> Decoded<User> {
     return curry(User.init)
-      <^> json <| "id"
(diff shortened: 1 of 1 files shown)


gfontenot:
Thanks for the PR!

I'm actually on the fence about merging this right now. We haven't released a new version with the new syntax and I haven't had time to spend on this lib so I'm not 100% sure when we will. The README currently reflects the most recent syntax available, which _seems_ right to me. I'd hate for people to come to the README to figure out how to use Argo and then be led astray, _especially_ since we don't have any concrete timeline for the next release.

PR #500, opened 2020-03-27
Migrate to Swift 5

juvs:
Migrate to Swift 5

### DIFF
--- a/Argo.podspec
+++ b/Argo.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
   spec.name = 'Argo'
-  spec.version = '4.1.2'
+  spec.version = '5.0.0'
   spec.summary = 'Functional JSON parsing library for Swift.'
   spec.homepage = 'https://github.com/thoughtbot/Argo'
   spec.license = { :type => 'MIT', :file => 'LICENSE' }
--- a/Argo.xcodeproj/project.pbxproj
+++ b/Argo.xcodeproj/project.pbxproj
@@ -776,7 +776,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0700;
-				LastUpgradeCheck = 0930;
+				LastUpgradeCheck = 1140;
 				ORGANIZATIONNAME = thoughtbot;
 				TargetAttributes = {
 					809754C01BADF34100C409E6 = {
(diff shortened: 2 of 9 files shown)


stefangiapantzakis:
@juvs Any idea if this will be merged soon?

gfontenot:
Hey, sorry about the delay on this! I haven't had nearly enough time to give this stuff the attention it deserves. I _think_ this all looks good with a couple of caveats:

1. `master` is, unfortunately, in a semi-finished state right now. I started a fairly major refactor a _long_ time ago and haven't had time to finish it up. So if the intent here is to get the shipping version onto Swift 5, we'll probably have to approach this differently. I still might merge this PR since it makes sense for `master` to be on Swift 5, but it's worth thinking about.
2. I'm _still_ unclear on how swift version bumps like this affect our versioning scheme. I think that the safest way forward here is probably to bump the shipping version to Swift 5 and then give it a major version bump? Does anyone have any thoughts/opinions on this?

stefangiapantzakis:
@gfontenot Hey! Thanks for such a quick response! Those are both good points.
It would be ideal if the shipping version supported Swift 5 just to make installation/development easier, but I understand this may not be as easy as merging this PR and creating a new release.

In terms of the versioning scheme, I was expecting any Swift 5 updates to cause a major version bump anyways.

What are my current options for using Argo on a brand new project targeting Swift 5?

gfontenot:
Ok so here's the plan:

1. I'm going to merge this (after verifying locally on master that no additional changes need to happen)
2. I'll update the current shipping version to 5.0 locally
3. I'll release that as v5.0

I'll post back here once all of that is done.

gfontenot:
Ugh, actually, this is going to take me a minute because I have to fix the Package.swift files as well for the new tool versions. I'll spend some time on this tonight though.

gfontenot:
Oh! Turns out: I've gone through this once already. Version 4.2 of Argo adds support for Swift 5 (via the Swift 4.2 tooling), so I think that should work fine? I'll still cut a new release based off that one that adds support for 5.0 and 5.2 explicitly, but it's good to know there's already a solution. I'm also going to close this PR in favor of a more comprehensive PR that updates our dependencies and adds better SPM support.

gfontenot:
Version 5.0 has been released which explicitly supports Swift 5, and I've also added these changes (along with some others) to master.

stefangiapantzakis:
@gfontenot Thanks mate! That's great news

stefangiapantzakis:
@gfontenot Will the CocoaPod Specs Repo be updated to point at 5.0? It seems it still points at 4.2.0

gfontenot:
Ah crap, sorry about that. It's up now!

